PR #1, opened 2018-03-26
Cognito authenticate

jbeard4:
Adds user authentication with cognito.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
     "test": "react-scripts test --env=jsdom",
     "eject": "react-scripts eject",
     "storybook": "start-storybook -p 9009 -s public",
-    "build-storybook": "build-storybook -s public"
+    "build-storybook": "build-storybook -s public",
+    "publish" : "aws s3 cp --recursive ./build s3://streetlives-web/ --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers"
   },
   "devDependencies": {
     "@storybook/addon-actions": "^3.3.15",
--- a/package.json
+++ b/package.json
@@ -3,6 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "aws-amplify": "^0.2.11",
+    "aws-amplify-react": "^0.1.36",
     "bootstrap": "^4.0.0",
     "classnames": "^2.2.5",
     "enzyme": "^3.3.0",
--- a/src/app/App.js
+++ b/src/app/App.js
@@ -1,11 +1,17 @@
 import React, { Component } from "react";
 import { BrowserRouter, Route, Switch } from "react-router-dom";
 
+import Amplify from 'aws-amplify';
+import { withAuthenticator } from 'aws-amplify-react';
+import aws_exports from './aws-exports';
+
 import Map from "./map/Map";
 import Login from "./login/Login";
 import Form from "./form/Form";
 
 import "./App.css";
+
+Amplify.configure(aws_exports);
 
 class App extends Component {
   render() {
@@ -23,4 +29,4 @@ class App extends Component {
   }
 }
 
-export default App;
+export default withAuthenticator(App);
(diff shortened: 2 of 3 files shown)


src/app/aws-exports.js
```diff
@@ -0,0 +1,10 @@
+export default {
+  Auth: {
+  // REQUIRED - Amazon Cognito Identity Pool ID
+      identityPoolId: 'us-east-2:a474fd65-5d53-4a95-8c88-62b800b8fbaf', 
+  // REQUIRED - Amazon Cognito Region
+      region: 'us-east-2',
+      userPoolId: 'us-east-2_1b7twTyIS',
+      userPoolWebClientId: '32qk7q3p62vpf85op2omg9v3n8'
```
nlaz: I'm assuming all these values are okay to be shared publicly?

src/app/App.js
```diff
@@ -1,20 +1,50 @@
...
 
 import "./App.css";
 
+Amplify.configure(aws_exports);
+
 class App extends Component {
+
+  componentWillMount(){
+    Amplify.Auth.currentAuthenticatedUser().then((user) => {
+      console.log('user',user);
```
nlaz: Do we want to keep this `console.log` in here?

src/app/App.js
```diff
@@ -1,20 +1,50 @@
...
+        attributes.forEach(x => o[x.Name] = x.Value)
+        this.setState({user : o});
+      })
+    });
+  }
+
+  logout() {
+    Amplify.Auth.signOut()
+        .then(data => {
+          window.location.href=window.location.href;
```
nlaz: Are you reloading the page here? I think you could use `window.location.reload()` instead.

nlaz:
I wasn't able to sign in. What am I doing wrong?

screen shot 2018-03-26 at 11 02 10 am

src/app/aws-exports.js
jbeard4: I think so. It's all information that must to be served to the client so that it can access AWS. None of it is secret.

src/app/App.js
jbeard4: I will remove.

src/app/App.js
jbeard4: Yes, they should be the same in the context. `window.location.reload()` is easier to read. I will replace it.

jbeard4:
I think it's two things:

1. It's sensitive about phone number format. Try appending "+1" to the number.
2. It's giving you a bogus error message when validation fails.

Please give it another try and see if it works.

package.json
```diff
@@ -19,7 +21,8 @@
     "test": "react-scripts test --env=jsdom",
     "eject": "react-scripts eject",
     "storybook": "start-storybook -p 9009 -s public",
-    "build-storybook": "build-storybook -s public"
+    "build-storybook": "build-storybook -s public",
+    "publish" : "aws s3 cp --recursive ./build s3://streetlives-web/ --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers"
```
Rovack: @jbeard4 Just to make sure: Versioning is enabled on this bucket, right?

src/app/aws-exports.js
Rovack: Just a thought, but do we prefer storing these values in the repo, or as environment variables or something in the build environment (i.e. currently our machines)?
There''s no security implication if they're client-side values anyway, just wouldn't necessarily want to have to make a commit if we just want to e.g. deploy to a different region.

If this is the standard way though, I'm sure this is fine.

package.json
jbeard4: I haven't enabled it. Do you think I should?

Rovack: If this is the bucket we intend to use for actual deployment of the app bundle, I think we should. That way we can easily rollback to a known working version if something fails.

jbeard4: OK, I'll add versioning. Thanks.

Rovack:
@jbeard4 Anything stopping us from merging this?
I'm working on a branch that's branched off this one, so it would be great if this can be merged before I open a PR for that one.

jbeard4:
@Rovack There are no blockers of which I am aware.

PR #2, opened 2018-03-26
Add variables to colors file

nlaz:
From @Rovack's feedback, I pulled out the css colors into a separate file.


Rovack:
LGTM, but maybe we could ask @noidsign about the palette UX have been using, and whether at least some of the values we have here have semantic meaning, so we can name them accordingly? (E.g. aliasing "yellow" to "primaryColor" or "secondaryColor" or whatever it is.)

PR #3, opened 2018-03-28
Add Google Analytics tracking to the routes

Rovack:


src/config.js
```diff
@@ -0,0 +1,3 @@
+export default {
+  gaTrackingId: process.env.GA_TRACKING_ID || "UA-116475051-1",
```
nlaz: CRA has a special way of dealing with [environment variables](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-custom-environment-variables). Looks like you should prefix the variable with `REACT_APP_` so in this case:`REACT_APP_GA_TRACKING_ID`.

nlaz: Because you don't like magic (🎉), here's the where it's handled in the ejected branch: https://github.com/streetlives/streetlives-web/blob/ejected_app/config/env.js#L61.

src/app/App.js
```diff
@@ -13,9 +14,9 @@ class App extends Component {
       <div className="App">
         <BrowserRouter>
           <Switch>
-            <Route exact path="/" component={Map} />
-            <Route path="/login" component={Login} />
-            <Route path="/form" component={Form} />
+            <Route exact path="/" component={withTracker(Map)} />
+            <Route path="/login" component={withTracker(Login)} />
+            <Route path="/form" component={withTracker(Form)} />
```
nlaz: Suggestion: This doesn't have to be an HOC and could exist as a standalone component which would avoid wrapping each component.

```jsx
import Tracker from "./Tracker";
// ...
<BrowserRouter>
   <Route path="/" component={Tracker} />
   <Switch>
       <Route exact path="/" component={Map} />
       <Route exact path="/" component={Login} />
       <Route exact path="/" component={Form} />
   </Switch>
</BrowserRouter>
```

src/config.js
Rovack: Great point; was not aware of it. Thanks!

src/app/App.js
Rovack: Yeah, I saw a bunch of discussions about this before going with the current solution.
It seems people use both approaches, but the HOC method is what appears in their official [demo](https://github.com/react-ga/react-ga/tree/master/demo) and [documentation](https://github.com/react-ga/react-ga/wiki/React-Router-v4-withTracker). Apparently, the reason it's preferred is that it gives you more flexibility in terms of tracking other components, beyond just the high-level routes. Having to repeat it in a few places didn't seem too bad to me, as long as it's mostly contained to this one file.

But, I don't feel very strongly about it either way. If you still think it's worth it, I can make the change.

src/config.js
Rovack: (Also, sorry: no idea how, but I didn't notice the comments until just now. Thank you for reviewing.)

src/app/App.js
nlaz: That's a good point. I don't mind either way. 👍

PR #4, opened 2018-03-29
Update eslint rules

nlaz:
This PR updates the eslint rules to follow the Airbnb style guide:`eslint-config-airbnb`. This matches with the config that we use on the streetlives-api repo with particular extensions for jsx code. The [Airbnb style guide rules](https://github.com/airbnb/javascript) are opinionated so it would be good to have some agreement on introducing these changes before merging.


package.json
```diff
@@ -7,23 +7,27 @@
...
   },
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test --env=jsdom",
     "eject": "react-scripts eject",
+    "lint": "eslint .",
     "storybook": "start-storybook -p 9009 -s public",
     "build-storybook": "build-storybook -s public"
-  },
```
Rovack: Why did these all have to move to normal dependencies?

src/app/App.test.jsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
```
Rovack: Wait, the _tests_ need to be ".jsx"? That seems like an odd rule...

.eslintignore
```diff
@@ -0,0 +1,3 @@
+dist/*
```
Rovack: Do we actually have a /dist directory here?

src/app/App.test.jsx
nlaz: Yeah. Any file with JSX in it needs a `.jsx` extension:

PR #5, opened 2018-04-09
Feature/map view

Rovack:
This adds a basic map page to the app, that shows all the locations in our database, with a few details about them shown when markers are clicked.
The behavior will need to be improved a lot (see https://trello.com/c/A5ztvqJA), but this is a start at least.

Note that some of the current changes here are actually not part of this feature, but just a result of master being merged into this branch but not into api-gateway, from which this one was branched off.
Once api-gateway is updated (and hopefully merged into master), this should all be clearer.


jbeard4:
LGTM

jbeard4:
@Rovack: @nlaz created a PR to merge feature/map-view directly into master, rather than api-gateway. We can therefore close this PR.

Rovack:
@jbeard4 @nlaz Ah, cool. In general I'd say changing the Base of an existing PR is usually preferable to opening a new one, just so we don't lose any comments, history etc, but doesn't really matter in the case of this one.

PR #6, opened 2018-04-14
`feature/map-view` -> `master`

nlaz:


Rovack:
@jbeard4 @nlaz Doesn't this fail lint? I thought e.g. double quotes were replaced with single ones, but files like App.jsx from map-view still use double quotes.

jbeard4:
We temporarily disabled lint.

PR #8, opened 2018-04-15
`feature/workflows` -> `master`

nlaz:


src/app/locationForm/location-name/LocationName.jsx
```diff
@@ -12,6 +13,14 @@ class LocationName extends Component {
     this.onConfirm = this.onConfirm.bind(this);
     this.onEdit = this.onEdit.bind(this);
     this.onSubmit = this.onSubmit.bind(this);
+
+    if (!props.locationData) {
+      props.getLocation(props.match.params.locationId);
+    }
+  }
+
+  componentWillReceiveProps(props){
```
Rovack: I'm not sure this is the behavior we want. Once the screen has rendered and the user may be interacting with it, suddenly changing modes automatically as data comes in may be weird, and also could create edge cases (e.g. in case the user has started typing something, it might just be wiped).

If we no longer want to assume the data was already fetched in a previous screen (which it currently always should be, except in development with live reload), then we should probably be blocking the UI with some indicator while it loads, rather than letting the user get into confusing situations.

src/app/locationForm/location-name/LocationName.jsx
```diff
@@ -48,16 +57,20 @@ class LocationName extends Component {
 }
 
 const mapStateToProps = (state, ownProps) => {
-  const { locationId } = ownProps.location.state;
+  const locationId = ownProps.match.params.locationId;
```
Rovack: I'm surprised eslint doesn't complain about this. On the back-end, at least, the airbnb rules enforce the slightly-more-ES7-idiomatic `const { locationId } = ownProps.match.params;`.

src/reducers/index.js
```diff
@@ -7,6 +7,30 @@ const dbReducer = (state = {}, action) => {
         return _.extend({ [action.payload.id] : action.payload}, state)
       }
       break;
+    case 'OPTIMISTIC_UPDATE_LOCATION':
+      if(action.payload){
+        return _.extend(
```
Rovack: Here too, looks like we could avoid underscore with something like:
```
return {
  ...state,
  [`last/${action.payload.id}`] : state[action.payload.id],
  [action.payload.id] : {
    ...state[action.payload.id],
    ...action.payload.params,
  },
};
```

src/reducers/index.js
```diff
@@ -7,6 +7,30 @@ const dbReducer = (state = {}, action) => {
...
+      }
+      break;
+    case 'ROLLBACK_UPDATE_LOCATION':
+      if(action.payload){
+        return _.extend(
+          {},
+          state,
+          {
+            [`last/${action.payload.id}`] : null,
+            [action.payload.id] : state[`last/${action.payload.id}`]
```
Rovack: Hmmm, implementing rollback in the reducer like this is interesting...
It certainly feels more Redux-y, though wouldn't this potentially lead to (probably rare) bugs when responses are slow, so the user can make more than 1 request before getting a response? (I mean, if e.g. 2 requests are made and then the first returns an error, it's actually the 2nd update that would get rolled back.)

Maybe it would be safer to leave rollback up to the thunk? The thunk could remember the exact fields that it changed and their old values, and in case of an error, dispatch an update with the old values for only those specific props.
Theoretically it could even remain the reducer's responsibility, but instead of keeping just 1 "last" value, it would store a mapping of request IDs to the changes they made.

Of course, even these approaches wouldn't help in the case both requests made changes to the same field.
Seems like handling that case would require maintaining a more complex history...

I would assume this is probably a common use case, so actually some source indicating what the best practice is would be great...
Googling briefly, it looks like https://github.com/ForbesLindesay/redux-optimist may have a request ID-based implementation ("transaction ID" technicallly) with a more complex history, so maybe that could work.

src/app/locationForm/common/FormView.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ConfirmationOptions from '../ConfirmationOptions';
```
Rovack: If we have a /common directory now, I'd think ConfirmationOptions should be in it too.

src/app/locationForm/common/FormEdit.jsx
```diff
@@ -0,0 +1,54 @@
+import React, { Component } from 'react';
```
Rovack: Whereas I think the Form container's logic is going to remain pretty standard across the different form screens, looking at the designs it seems like the presentation is actually going to be fundamentally different for nearly every screen (other than probably description and location name). So, I think we'll need the flexibility to specify them separately, which FormView and FormEdit can't really provide.
Of course, I don't think there's any harm in having FormView and FormEdit for convenience right now, before we build out the actual screens.

src/app/locationForm/additional-info/AdditionalInfo.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Header from '../../../components/header';
+import Button from '../../../components/button';
+import Input from '../../../components/input';
+
+function LocationDescription(props) {
```
Rovack: Function name needs to be updated.

src/app/locationForm/description/LocationDescription.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
```
Rovack: It looks like this is actually supposed to be the organization description, not location description.
In the most recent designs, at least, the text is "How would you describe this organization?". (And also, in our initial data we only have descriptions of organization, no locations.)

src/reducers/index.js
```diff
@@ -0,0 +1,23 @@
+import _ from 'underscore';
+
+const dbReducer = (state = {}, action) => {
```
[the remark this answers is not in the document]
jbeard4: Fixed in d413d0fae7103d5dd5129cc348781c7729cec938

src/app/locationForm/common/FormView.jsx
jbeard4: Should be fixed in 2398171c5501714a4ace402c79c4aa518cae2e2e

src/app/locationForm/additional-info/AdditionalInfo.jsx
jbeard4: Should be fixed in a969f30

src/app/locationForm/description/LocationDescription.jsx
jbeard4: Fixed in 8383dbc

src/app/locationForm/common/FormEdit.jsx
jbeard4: Yes, I agree, most edit forms are going to be pretty custom, and different from other forms. Please see my solution in https://github.com/streetlives/streetlives-web/pull/8/commits/8723b4ce0c82ccd767c5542d324e568d3635d336#diff-c57e579bad4f4f74984ee84e76c54e7dR14 which is to create a custom LocationNumberEdit form.

src/app/locationForm/number/LocationNumber.jsx
```diff
@@ -33,17 +121,24 @@ const mapStateToProps = (state, ownProps) => {
...
-              locationData.Phones[0] && 
-              locationData.Phones[0].number ? 
-              locationData.Phones[0].number : null,
+    value : {
+      number: locationData && 
+                locationData.Phones &&
+                locationData.Phones[0] && 
+                locationData.Phones[0].number ? 
+                locationData.Phones[0].number : null,
+      extension: locationData &&
```
Rovack: Doesn't seem necessary to repeat this set of checks for both number and extension - I think this could just be something along the lines of:
```
const phone = locationData && locationData.Phones && locationData.Phones[0];
const phoneFields = phone ? { number: phone.number, extension: phone.extension } : {};
return {
  value: phoneFields,
  locationData,
};
```

(Honestly, even the filtering of fields is probably unnecessary. The whole thing could probably just be `value: locationData && locationData.Phones && locationData.Phones[0]`.)

src/app/locationForm/common/Form.jsx
```diff
@@ -0,0 +1,61 @@
+import React, { Component } from 'react';
```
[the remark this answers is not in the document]
jbeard4: I updated the login in `isEditing`, and generalized `Form.value` so that it can take an object of multiple fields. Please see my solution in https://github.com/streetlives/streetlives-web/pull/8/commits/8723b4ce0c82ccd767c5542d324e568d3635d336#diff-71061816a970b190e63f516087e7325dR18. This is for the Phone Number form, which takes a phone number and extension. Furthermore, it splits up the phone number into three different fields for area code, first three digits, and last four digits.

You can see how these fields get passed back to dispatch in the Edit form here:
https://github.com/streetlives/streetlives-web/pull/8/commits/8723b4ce0c82ccd767c5542d324e568d3635d336#diff-c57e579bad4f4f74984ee84e76c54e7dR55

And dispatched to the reducer here: 
https://github.com/streetlives/streetlives-web/pull/8/commits/8723b4ce0c82ccd767c5542d324e568d3635d336#diff-c57e579bad4f4f74984ee84e76c54e7dR141

Please let me know what you think. Thanks.

src/app/locationForm/number/LocationNumber.jsx
```diff
@@ -33,17 +121,24 @@ const mapStateToProps = (state, ownProps) => {
   const locationData = selectLocationData(state, locationId);
 
   return {
-    value: locationData && 
-              locationData.Phones &&
-              locationData.Phones[0] && 
-              locationData.Phones[0].number ? 
-              locationData.Phones[0].number : null,
+    value : {
+      number: locationData &&
```
Rovack: According to the designs the `description` field should also be supported.

src/app/locationForm/number/LocationNumber.jsx
```diff
@@ -1,25 +1,113 @@
...
+  }
+}
 
 const LocationNumberView = compose(
   withProps({
     topText : 'LOCATION PHONE NUMBER',
   })
-)(props => <FormView {...props} />)
+)(props => {
+  const value = props.value && props.value.number && `${props.value.number.replace(/\./g,'-')}${props.value.extension ? ` ext.${props.value.extension}` : ''}`;
```
Rovack: I assume this is just a temporary way of showing the number, and later we'll make it match the designs? (And also add support for the list of numbers shown in the [designs](https://www.figma.com/file/fdzJjvkSugVGnUa3zNdvMNsU/StreetTeam_source_of_truth?node-id=4%3A1290)?)

Rovack: At that point I suppose we'll also need the LocationNumberView to be a custom component.
That makes me think it probably makes sense to break this file down to split the container and presentational components, as even with just LocationNumberEdit inside it it's already doing quite a lot, and probably moreso once we have a proper LocationNumberView.

src/reducers/index.js
jbeard4: All set

src/app/locationForm/number/LocationNumber.jsx
Rovack: And of course, a lot of util libraries just have functions for getting a nested property like this with 1 line (e.g. lodash's `get`). I guess the closest thing underscore has is `_.property`, though...

src/app/locationForm/number/LocationNumber.jsx
```diff
@@ -1,25 +1,113 @@
...
+      extension : props.value.extension || ''
+    };
+
+    this.onSubmit = this.onSubmit.bind(this);
+    Object.keys(this.state).forEach(k => {
+      this[`onChange_${k}`] = this.onChange.bind(this, k);
+    });
+  }
+
+  componentWillReceiveProps(props){
```
Rovack: Here, too, once the user has started entering data, I don't think we'll want to overwrite it even if the server data changes, so componentWillReceiveProps should probably be removed.

(And if we did want the server updates to overwrite user input, probably the new getDerivedStateFromProps would help avoid all this code duplication.)

src/app/locationForm/common/Form.jsx
```diff
@@ -15,8 +15,18 @@ class Form extends Component {
     }
   }
 
+  isEditing(props){
+    return props.value === null ||  
+      ( typeof props.value === 'object' ?
```
Rovack: If we want to keep the logic of whether a form is empty standardized like this in the HOC, we'll probably need to add more cases to it. For instance, once Phones supports an array, a new case for arrays will need to be added here.

As mentioned in another comment, IMO it would be more flexible to keep the HOC simple, and let the specific forms determine whether a given form is empty or not (letting the HOC know through a "isFormEmpty" prop or whatnot), even if it means a line or two of duplication in some of the forms that really do only have 1 simple value.

src/app/locationForm/location-name/LocationName.jsx
jbeard4: Should be fixed in 4565fa5442acafdc7ac23a46bf862b563dd7391e

src/app/locationForm/common/Form.jsx
```diff
@@ -15,8 +15,18 @@ class Form extends Component {
     }
   }
 
+  isEditing(props){
+    return props.value === null ||  
+      ( typeof props.value === 'object' ? 
+          Object.keys(props.value).some( key => !props.value[key] ) :
```
Rovack: IIUC, we will go into edit mode if even _one_ of the fields is empty.
Seeing as some fields may be optional and never receive a value (e,g. extension), I think it should be the other way around: we go into edit mode only if _none_ of the fields have a value.
We could verify with UX.

src/app/locationForm/common/Form.jsx
```diff
@@ -15,8 +15,18 @@ class Form extends Component {
...
+          !props.value
+      );
+  }
+
   componentWillReceiveProps(props){
-    this.setState({ isEditing: !props.value });
+    this.setState({ 
+      isEditing: this.isEditing(props)
```
Rovack: Same comment about componentWillReceiveProps noted elsewhere.

src/app/locationForm/location-name/LocationName.jsx
jbeard4: Here's what it looks like:

screen shot 2018-04-21 at 9 32 55 pm
screen shot 2018-04-21 at 9 27 22 pm

src/app/locationForm/number/LocationNumber.jsx
Rovack: That would also mean we probably don't need to add the `value` props to `Input`, and can just use `defaultValue`.

src/reducers/index.js
```diff
@@ -1,37 +1,4 @@
...
-      break;
-    default:
-      return state;
-  }
-
-  return state;
-}
+import dbReducer from './db';
 
 export const selectLocationData = (state, locationId) => state.db[locationId];
```
Rovack: This probably belongs in the db reducer.

src/app/locationForm/description/OrganizationDescription.jsx
```diff
@@ -8,23 +8,23 @@ import FormEdit from '../common/FormEdit';
...
-    headerText : 'Please describe this location',
+    headerText : 'How would you describe this organization?',
     placeholderText : 'Enter a description of the location',
   })
 )(props => <FormEdit {...props} />)
 
-const LocationDescriptionView = compose(
+const OrganizationDescriptionView = compose(
   withProps({
     topText : 'LOCATION DESCRIPTION',
```
Rovack: The topText and placeholderText are still the old ones for location description.

src/app/locationForm/common/Form.jsx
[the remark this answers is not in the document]
Rovack: Yeah, I think this solution should work.
I left a comment on the Phone comment, mentioning I still think it might be more flexible to have the specific forms pass an explicit prop (so Form doesn't have to care about the values at all), but functionally either approach should work.

src/app/locationForm/common/Form.jsx
```diff
@@ -42,6 +43,8 @@ class Form extends Component {
   }
 
   render() {
+    if (!this.props.locationData) return <LoadingLabel />;
```
Rovack: Ah, it's really just one line. Awesome!

src/reducers/index.js
Rovack: Yeah, you're right. We probably shouldn't do the thunk option.
I think redux-optimistic is the way to go.

src/app/locationForm/location-name/LocationName.jsx
jbeard4: eslint passes, and doesn't complain. I'm not sure why.

src/app/locationForm/common/Form.jsx
jbeard4: 👍

src/reducers/index.js
jbeard4: Fixed in b5a14de702a74df30597583ec35db824963a009a

src/app/locationForm/description/OrganizationDescription.jsx
jbeard4: Thanks, I didn't see that. Fixed in 24f4b92d804edef6fca11bc2b9baaf9254c8d56a

src/reducers/index.js
Rovack: I mean, thunks do have access to the state (they get passed both dispatch and getState), but the thunk approach still wouldn't solve the case where it's the same field being updated multiple times, and regardless I think it's just better to go with an established solution rather than build our own.

src/components/map/LocationMarker.jsx
```diff
@@ -1,22 +1,24 @@
...
 class LocationMarker extends Component {
-
-  constructor(props){
+  constructor(props) {
     super(props);
     this.handleYesClick = this.handleYesClick.bind(this);
+    this.onToggleInfo = this.onToggleInfo.bind(this);
   }
 
-  onToggleInfo = () => {
```
Rovack: Ah right, eslint on streetlives-web doesn't like class properties for some reason... I wonder why that is; it doesn't mind them on streetlive-api...

src/app/locationForm/number/LocationNumber.jsx
```diff
@@ -1,25 +1,113 @@
...
+    });
+    this.props.onSubmit();
+  }
+
+  render() {
+    return (
+      <div>
+        <Header>What's this location's phone number?</Header>
+          <div>
+            <Input
```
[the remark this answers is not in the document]
jbeard4: Fixed in 2fcfb57fa0d98863861d0abda8ab2c12766a602e

src/app/locationForm/number/LocationNumber.jsx
```diff
@@ -1,25 +1,113 @@
...
+      extension : this.state.extension
+    });
+    this.props.onSubmit();
+  }
+
+  render() {
+    return (
+      <div>
+        <Header>What's this location's phone number?</Header>
+          <div>
```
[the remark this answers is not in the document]
jbeard4: What issues are you seeing in indentation?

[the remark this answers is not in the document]
Rovack: <div> and <Header> are on the same level, but it looks like <div> is more indented.

src/app/locationForm/number/LocationNumber.jsx
```diff
@@ -1,25 +1,113 @@
...
+            <Input
+              style={{width:'2em'}}
+              type="number"
+              value={this.state.firstThree}
+              onChange={this.onChange_firstThree}
+            />&nbsp;-&nbsp;
+            <Input
+              style={{width:'3em'}}
+              type="number"
+              value={this.state.lastFour}
```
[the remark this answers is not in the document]
jbeard4: I'm looking into using HTM5 form validation. It seems to be well-supported across browsers:

https://caniuse.com/#feat=form-validation

src/app/locationForm/number/LocationNumber.jsx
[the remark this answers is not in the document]
jbeard4: Oh, nevermind, I see the problem now. Yes, it's strange that eslint didn't complain about this.

src/app/locationForm/number/LocationNumber.jsx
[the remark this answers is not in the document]
jbeard4: Fixed in 2c0cd11d0215cc55b7d10c6f6ec226ce02f2e9c1

Preview:

screen shot 2018-04-22 at 12 49 33 pm
screen shot 2018-04-22 at 12 49 18 pm
screen shot 2018-04-22 at 12 48 56 pm

src/app/locationForm/number/LocationNumber.jsx
jbeard4: Fixed in 85bab537c37f37de9a01b31aead81da23c950c40 per your suggestion.

src/app/locationForm/number/LocationNumber.jsx
jbeard4: Yes, exactly, this is a temporary view until we can make it better match the designs.

src/app/locationForm/common/Form.jsx
jbeard4: I think this is an appropriate use of componentWillReceiveProps. I don't think we are likely to lose data here, because we will not render the form UI until the data is loaded. But we need to initialize the state.isEditing when the data is loaded, and it would make sense to do this in componentWillReceiveProps. After state.isEditing is set, then the form will be displayed, and the user can edit data.

src/app/locationForm/common/Form.jsx
jbeard4: Fixed in a463b359a697bd181e1ff8fedf18d39e65afc03a

src/app/locationForm/common/Form.jsx
jbeard4: I think it would make sense to pass this logic into the Form HOC as a prop whose value is a function. I'll add this prop as soon as we require it.

src/app/locationForm/number/LocationNumber.jsx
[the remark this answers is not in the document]
Rovack: Nice!

src/app/locationForm/number/LocationNumber.jsx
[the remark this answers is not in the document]
Rovack: Ha, never realized GitHub parses HTML... I meant to say "div and Header are on the same level, but it looks like div is more indented".

src/app/locationForm/common/Form.jsx
Rovack: Hmm, I was thinking it would be possible to do it even without a function. Since e.g. LocationForm controls the mapStateToProps which gives the Form its props, it could just determine the value there and pass it, without Form needing to be passed any function.
And yeah, some default could make sense either way. Form could fall back to checking `value` if no `isEmpty` prop is passed, or something like that.

src/app/locationForm/common/Form.jsx
Rovack: Yeah, you're right. I guess we have to support changing the value after the component mounts, since it may need the time to load...

It almost feels like the user-entered state should be separate from the prop, and render should decide which of the two to use: rendering with the value from the prop as long as the user-entered state doesn't exist yet, and then switching to rendering with the value from state.
But, that sounds more complicated.

Another option would be to remove the component state entirely, and manage it all in Redux state, which is what I've seen done in a lot of apps. That way the component can always just render the props, and on change dispatch an action. Submitting, at that point, is just a matter of calling the API with the values from Redux state.
But that, again, sounds like a bigger change.

Anyway, as you say, it's probably unlikely anything will actually update the state after the user has started typing, at least the way the rest of the app works right now (with no push updates from the server, no background calls, or anything like that), so we should be fine for now.

src/app/locationForm/common/Form.jsx
```diff
@@ -19,7 +19,7 @@ class Form extends Component {
   isEditing(props){
     return props.value === null ||  
       ( typeof props.value === 'object' ? 
-          Object.keys(props.value).some( key => !props.value[key] ) :
+          Object.keys(props.value).some( key => props.value[key] ) :
```
Rovack: Now that I think about it, when I said "other way around", that was pretty confusing. :P
I'm pretty sure this way, we'll actually enter editing mode if we _do_ have at least one value, which is exactly the case where we should be in viewing mode.

What I meant is that I think we should enter editing mode if none of the fields have a value, i.e. changing from `some(!value)`->`every(!value)`, not `some(!value)`->`some(value)`.

jbeard4: Fixed in 23a8250468da62b007a56a42c5d65baf6a0cefe7

src/reducers/db.js
```diff
@@ -32,6 +33,24 @@ export const dbReducer = (state = {}, action) => {
         };
       }
       break;
+    case OPTIMISTIC_UPDATE_PHONE:
+      const location = state[action.payload.locationId];
+      const idx = location.Phones.map( phone => phone.id).indexOf(action.payload.phoneId)
```
Rovack: findIndex could save an iteration here.

src/reducers/db.js
```diff
@@ -32,6 +33,24 @@ export const dbReducer = (state = {}, action) => {
...
+      const phone = location.Phones[idx]
+      const newPhones = location.Phones.slice(0,idx).concat({...phone, ...action.payload.params}).concat(location.Phones.slice(idx+1))
```
Rovack: I think this could be:
```
const newPhones = [...location.Phones.slice(0, idx), {...phone, ...action.payload.params}, ...location.Phones.slice(idx+1)];
```

src/reducers/db.js
```diff
@@ -32,6 +33,24 @@ export const dbReducer = (state = {}, action) => {
...
+      console.log('newPhones',newPhones);
```
Rovack: Probably not needed for production, right?

src/reducers/db.js
```diff
@@ -32,6 +33,24 @@ export const dbReducer = (state = {}, action) => {
...
+      if (action.payload) {
+        return {
+          ...state,
+          [`last/${action.payload.locationId}`]: state[action.payload.locationId],
+          [action.payload.locationId]: {
+            ...state[action.payload.locationId],
+            ...action.payload.params,
```
Rovack: Wait, the params are for the phone, not the location, right?

src/app/locationForm/number/LocationNumber.jsx
```diff
@@ -42,7 +42,7 @@ class LocationNumberEdit extends Component {
   }
 
   onSubmit() {
-    this.props.updateValue({
+    this.props.updateValue(this.props.value.id, {
```
Rovack: If the location doesn't already have a phone, won't this produce a `Cannot read property 'id' of undefined` error?

src/reducers/db.js
jbeard4: Good catch

src/reducers/db.js
jbeard4: Fixed in cdf652af033e7b9e18f52372d76b55de2ae0efb7

src/reducers/db.js
jbeard4: Fixed in cdf652af033e7b9e18f52372d76b55de2ae0efb7

src/reducers/db.js
jbeard4: Fixed in cdf652af033e7b9e18f52372d76b55de2ae0efb7

src/app/locationForm/number/LocationNumber.jsx
jbeard4: Fixed in 3ffaa7cfd008343d6c966f10d3416ef4d943f154. Thanks.

jbeard4: Tested on location 04dcd3a8-5ed0-4d2a-a357-56b0703f3c8d, which does not have a phone number set.

src/services/api.js
```diff
@@ -42,19 +42,25 @@ export const getLocation = ({ id }) =>
...
         Authorization: idJwtToken,
       },
     });
   });
 
-export const updateLocation = updateResource.bind(this, 'locations');
-export const updatePhone = updateResource.bind(this, 'phones');
+export const updateLocation = updateResource.bind(this, { pathPrefix: 'locations', method: 'patch' });
+export const updatePhone = updateResource.bind(this, { pathPrefix: 'phones', method: 'patch' });
+export const createPhone = updateResource.bind(this, { pathPrefix: 'locations', method: 'post', pathSuffix: 'phones' });
```
Rovack: Personally I would probably favor simplicity over DRYness in this case (as we're getting close to the point `updateResource` will need to take as many args as `axios.request` :P), but doesn't make much difference either way.

PR #9, opened 2018-04-17
Support different heading types

nlaz:
Update `Header` component with different heading types.

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -1,5 +1,40 @@
+@value darkerGray from "./colors.css";
+
 body {
   margin: 0;
   padding: 0;
-  font-family: sans-serif;
+  font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
+}
+
+h1 {
+  font-style: normal;
+  font-weight: bold;
+  line-height: 28px;
+  font-size: 21px;
+  letter-spacing: -0.741176px;
+  color: darkerGray;
+}
+h2 {
+  font-weight: bold;
+  line-height: 24px;
+  font-size: 17px;
+  letter-spacing: -0.741176px;
+  color: darkerGray;
+}
+h3 {
+  font-style: normal;
+  font-weight: normal;
+  line-height: 24px;
+  font-size: 17px;
+  letter-spacing: -0.370588px;
+  color: darkerGray;
+}
+
+p {
+  font-style: normal;
+  font-weight: normal;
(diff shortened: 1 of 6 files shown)


Rovack:
Where are the styles for h1-h6 specified? Or are we just using the bootstrap headers?
If so, do they match the UX designs?

nlaz:
@Rovack The default h1-h6 styles are specified by bootstrap. I believe they are close to the designs. Let me if there are changes you need.

Rovack:
Ah alright, if we think they're close enough, then I don't have any other comments on this. Looks good.

Rovack:
Oh actually @nlaz, one more thing I remembered: There are some screens that could use this new version of Header, which previously had to use raw divs with custom styles.
I know Location Name is like that, but there might also be others. Probably would be good to update them now to use this component.

nlaz:
@Rovack @noidsign I updated the `h1`, `h2`, `h3`, and `p` global styles. I believe the header style on the location flow doesn't match any styles on the UI component list:
screen shot 2018-04-24 at 7 04 48 pm
screen shot 2018-04-24 at 7 05 32 pm

We should probably treat those headers differently in the meantime and push these changes forward. How does that sound?

noidsign:
@nlaz 
+ @Rovack  FYI
The component for the title highlighted in the location flow screenshot is ```h4```  from the UI Components page.

src/index.css
```diff
@@ -1,5 +1,40 @@
+@value darkerGray from "./colors.css";
+
 body {
   margin: 0;
   padding: 0;
-  font-family: sans-serif;
+  font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
+}
+
+h1 {
```
Rovack: Not that it matters too much, but do we not want these 3 in the CSS file for the Header component?

nlaz: They shouldn't be in the Header component because if someone uses `h1` without the component it should still have the same style.

Rovack: That's actually exactly what I was wondering about.
I would imagine generally no-one should be using raw tags, but if they do use e.g. a raw h1 header, bypassing our component, that would mean they don't want the standardization from the style guide. Just like if someone uses a raw input tag, bypassing our Input, they wouldn't get any of our styles for that.
But, I'm not too sure either way.

PR #11, opened 2018-04-19
Update location list style

nlaz:
screen shot 2018-04-19 at 7 00 23 pm


src/app/locationInfo/LocationField.jsx
```diff
@@ -19,34 +18,42 @@ class LocationField extends Component {
...
+    const { title, updatedAt } = this.props;
 
     const classNames = {
-      field: cx('LocationField border-top border-bottom', { 'bg-yellow': required }),
-      updatedAt: cx({
-        'text-danger': updatedAt && moment(updatedAt).isSameOrBefore(moment().subtract(1, 'years')),
-        'text-warning': updatedAt && moment(updatedAt).isSameOrBefore(moment()),
+      field: cx('LocationField border-top border-bottom'),
+      emptyDot: cx('rounded-circle mr-2', {
+        'bg-danger': !updatedAt || moment(updatedAt).isBefore(moment().subtract(1, 'years')),
```
Rovack: Since we're likely to need to iterate on the specifics of this logic quite a bit, I think it would make sense to extract the logic, or at least the values, so they're not just in the middle of a component.
Something as simple as putting the values in a constants file of some sort (maybe even the config) would probably suffice.

nlaz: Added some helper methods here: https://github.com/streetlives/streetlives-web/pull/11/files#diff-767d3777ac6f89301aa7bd4a3c65e3f2

PR #12, opened 2018-04-29
Creating initial services container

nlaz:
### Description
This PR sets up the initial container and interface for the services flow. It creates the container which has temporary list of service categories and a temporary list of services. It also handles the transition between the services info and service recap pages.

### Ticket
Related to https://trello.com/c/nv27FfX7/94-ui-build-service-flow-container

### Screenshots
screen shot 2018-04-29 at 2 17 44 pm
screen shot 2018-04-29 at 2 18 20 pm


src/app/App.jsx
```diff
@@ -34,6 +35,7 @@ function App({ authState }) {
             <Route path="/recap/:locationId" component={withTracker(Recap)} />
             <Route path="/location/:locationId" component={withTracker(LocationInfo)} />
             <Route path="/questions/:id/:locationId" component={withTracker(LocationForm)} />
+            <Route path="/services" component={withTracker(ServiceContainer)} />
```
jbeard4: @nlaz  You will probably want to include the locationId in the /services route so that on page refresh you can continue working on the services flow for that location.

src/app/services/ServiceContainer.jsx
```diff
@@ -0,0 +1,63 @@
...
+  },
+  other: {
+    1: { name: 'Project ID' },
+  },
+};
+
+const PAGES = {
+  INFO: 'Services info',
+  RECAP: 'Services recap',
+};
```
jbeard4: This stuff should also probably go into the route so that when the user refreshes the page, they can continue in their existing flow.

jbeard4: The location flow can probably be used as a reference for this.

nlaz: Following up from our offline conversation, we decided to use the server to persistent the data between these screens. I'll update this branch to mock these endpoints until we have them.

src/app/App.jsx
Rovack: Yeah, it would probably make sense to actually structure the front-end URLs kinda like REST, e.g.:
```
<Route path="/location/:locationId" component={withTracker(LocationInfo)} />
<Route path="/location/:locationId/recap" component={withTracker(Recap)} />
<Route path="/location/:locationId/questions/:id" component={withTracker(LocationForm)} />
<Route path="/location/:locationId/services" component={withTracker(ServiceCategoriesContainer)} />
<Route path="/location/:locationId/services/recap" component={withTracker(ServiceForm)} />
<Route path="/location/:locationId/services/:serviceId" component={withTracker(ServiceForm)} />
```
etc.

.eslintrc.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
     'no-console': 2,
     'no-multiple-empty-lines': [2, { max: 1, maxBOF: 1, maxEOF: 1 }],
     'no-unused-vars': [2, { args: 'none' }],
+    'react/prop-types': 0,
```
Rovack: Finally. :P

src/app/services/ServiceContainer.jsx
```diff
@@ -0,0 +1,63 @@
+import React, { Component } from 'react';
+import Button from '../../components/button';
+import NavBar from '../NavBar';
+import ServicesInfoPage from './ServicesInfoPage';
+import ServicesRecapPage from './ServicesRecapPage';
+
+const data = {
```
Rovack: You should be able to use the /taxonomy endpoint for this now instead of hard-coding.

Granted, it's still showing the categories from the NYPL PDF, but I think that's fine. At the very least, it'll be a very strong reminder that we need to map those services already. :sweat_smile:

src/app/services/ServiceContainer.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+  onNext = () => {
+    if (this.state.page === PAGES.INFO) {
+      this.setState({ page: PAGES.RECAP });
+    }
+  };
+
+  onSelectService = (service, id) => {
+    const { services } = this.state;
+    services[service][id].selected = !services[service][id].selected;
```
Rovack: I assume this is where we'll need to have some kind of check to prevent toggling services that already exist in the back-end, if we want that?

src/app/services/ServiceContainer.jsx
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  render() {
+    const { page, services } = this.state;
+    return (
+      <div className="text-left">
+        <NavBar title={PAGES.INFO} />
+        <div className="mb-5">
+          {page === PAGES.INFO && (
+            <ServicesInfoPage services={services} onSelectService={this.onSelectService} />
```
Rovack: I assume this is where we'll need to have some kind of check to prevent toggling services that already exist in the back-end, if we want that?

src/app/services/ServiceContainer.jsx
```diff
@@ -0,0 +1,63 @@
...
+          )}
+          {page === PAGES.RECAP && <ServicesRecapPage services={services} />}
+        </div>
+        <div className="position-fixed" style={{ right: 0, bottom: 0, left: 0 }}>
+          <Button fluid primary onClick={this.onNext}>
```
Rovack: Looking at onNext, it seems it only does anything in INFO mode.
So, shouldn't this button only appear in INFO mode?
(I may just be misunderstanding what this button does.)

src/app/services/ServicesRecapPage.jsx
```diff
@@ -0,0 +1,44 @@
...
+  getServiceItems = (category) => {
+    const { services } = this.props;
+
+    return Object.keys(services[category]).map((id) => {
+      const service = services[category][id];
+
+      return service.selected ? (
+        <ListItem key={id} title={service.name} progress={service.progress || 0} />
+      ) : (
+        undefined
```
Rovack: It at least used to be that React would only agree to render null, but not undefined.
I guess that might have changed though.

src/app/services/ListItem.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import Icon from '../../components/icon';
+
+function ListItem({ title, progress, onClick }) {
+  return (
+    <div className="ListItem py-4 border-top border-bottom" style={{ cursor: 'pointer' }}>
```
Rovack: I assume these will become actual links in a later PR, right?

nlaz: Fixed in this one 👍

src/app/services/ServicesRecapPage.jsx
nlaz: Code was removed with changes

src/app/services/ServicesInfoPage.jsx
```diff
@@ -0,0 +1,90 @@
...
+  render() {
+    const { active } = this.state;
+    const { services } = this.props;
+
+    return (
+      <div className="text-left">
+        <div className="py-5 px-3 container">
+          <Header>What programs and services are available at this location?</Header>
+        </div>
+        <Accordion>
```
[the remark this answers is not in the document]
nlaz: The accordions aren't iterating on the data but simply renders any child content. Maybe I'm misunderstanding the question.

src/app/services/ServiceContainer.jsx
nlaz: From the designs, I'm not too sure what this button does. Let's punt on this and follow up in another ticket.

Rovack: 👍

src/app/services/ServicesInfoPage.jsx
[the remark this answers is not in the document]
Rovack: Ha, right. I meant not the accordion components themselves, but rather what's being passed to them here, in ServicesInfoPage. :)
At least at the time I was making the comment, it was hard-coded rather than being based on the category data.

src/app/services/ServicesRecapPage.jsx
Rovack: I thought I still saw it somewhere else (maybe one of the other PRs?), though either way it makes little difference.

src/app/service/categories/ServiceCategories.jsx
```diff
@@ -0,0 +1,105 @@
...
+            <Header>What programs and services are available at this location?</Header>
+          </div>
+          <Accordion>
+            <Accordion.Item
+              active={active === 0}
+              onClick={() => this.onToggleOpen(0)}
+              title="All Services"
+              icon="home"
+            />
+            <Accordion.Content active={active === 0}>
```
Rovack: I'm not entirely sure I follow what's happening here.
IIUC, there's only 1 accordion, with a selector containing an option for each taxonomy item at the first level? Isn't it supposed to be 1 accordion for each taxonomy item at the first level, with a selector containing options for each of its child taxonomies? Currently, I don't see where `category.children` are used.

Also, I'd expect that rather than "Add another shelter service", we'd have something like `Add another ${category.name} service`, as I don't recall anything special about shelter specifically in the designs.

Rovack: Btw, if you're testing this with a local server (and not the dev RDS), it might help to run `DELETE FROM taxonomies;` and `DELETE FROM services;`, and then `sequelize db:seed:all` in streetlives-api (assuming you have the sequelize CLI installed).
That would seed the Streetlives taxonomy in the DB, instead of the Connections NYPL one.

nlaz: Ah understood. I wasn't seeing this updated data so I didn't know how best to render it. I followed your steps and updated this PR to dynamically render this data.

src/app/service/categories/ServiceCategories.jsx
```diff
@@ -61,26 +69,32 @@ class ServiceCategories extends Component {
...
-                    onClick={() => this.onSelect(category)}
-                    active={selected[category.id]}
-                  >
-                    {category.name}
-                  </Selector.Option>
-                ))}
-                <Selector.Option align="center">+ Add another shelter service</Selector.Option>
-              </Selector>
-            </Accordion.Content>
+            {taxonomy.map((category, i) => (
```
Rovack: Probably makes more sense for active to be based on the category ID, rather than the index, no?

src/app/service/categories/ServiceCategories.jsx
```diff
@@ -61,26 +69,32 @@ class ServiceCategories extends Component {
...
+              <div key={category.id}>
+                <Accordion.Item
+                  active={active === i}
```
Rovack: Probably makes more sense for `active` to be based on the category ID, rather than the index, no?

src/app/service/categories/ServiceCategories.jsx
```diff
@@ -61,26 +69,32 @@ class ServiceCategories extends Component {
...
+                  onClick={() => this.onToggleOpen(i)}
+                  title={category.name}
+                  icon={iconNames[category.name]}
+                />
+                <Accordion.Content active={active === i}>
+                  <Selector fluid>
+                    {category.children.map(service => (
```
Rovack: The name "service" might be a bit confusing here (the actual "service" object is separate, and is what the next screens are about), I'd go with something like "subcategory".
Then again, our terminology in the team outside Tech often does refer to these subcategories as "services", so maybe this is fine. I dunno.

src/app/service/recap/ServiceRecap.jsx
```diff
@@ -8,6 +8,12 @@ import SectionHeader from '../../../components/sectionHeader';
 import NavBar from '../../NavBar';
 import ListItem from './ListItem';
 
+const iconNames = {
```
Rovack: Could this be moved to a util or something, so it can be shared between ServiceCategories and this?

PR #14, opened 2018-04-30
Feature/workflows+usability fixes

jbeard4:
* Fix bug where width and height of map InfoBox exceed the boundaries of small screen sizes.
* Fixed layout bug on recap page, which was causing the whole page to scroll, rather than just the recap text.
* Added cancel button to edit forms
* Added code to disable auth (e.g. for when you are working on an airplane and can't reach AWS
* Added support for Thank You screen after finishing Location edit flow.


jbeard4:
I think this is showing lots of commits because the other branch got squashed.

src/app/locationForm/common/Form.jsx
```diff
@@ -42,13 +43,17 @@ class Form extends Component {
     this.setState({ isEditing: false }, this.props.onFieldVerified);
   }
 
+  onCancel() {
+    this.setState({ isEditing: false });
```
Rovack: I'm not really sure about this, but cancelling will unmount the editComponent, won't it?
If so, I would assume the state of any edits the user has started will be wiped.

If this is true, I think we should either add a confirmation, or somehow try to preserve the state (e.g. by storing it above the editComponent, or just making it always render but sometimes be invisible, etc).
Otherwise, I can only imagine the frustration a Street Team member will feel if they type a bunch of stuff, try to hit OK but have their finger slip and hit Cancel, and lose everything they've done...

src/app/locationForm/thanks/heart.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
```
Rovack: Ha, this isn't just an image? :)

src/app/locationForm/common/Form.jsx
jbeard4: Noted. I will add a confirmation

src/app/locationForm/thanks/heart.jsx
jbeard4: It's an SVG. We are passing width/height through as params, so we still need React ;)

src/app/locationForm/common/Form.jsx
noidsign: @Rovack @jbeard4 At this point, I would recommend *not* to add a confirmation screen for canceling. The use case imagined by @Rovack is mitigated by three elements: 1. small enough units to edit, 2. layout of cancel / OK buttons, and 3. when the cancel button is actually shown to users. It's more of a pain for users who need to cancel something having to go through a confirmation screen. Please do not add a confirmation screen for canceling at this point. Thanks.

noidsign: @jbeard4 @Rovack (Was still typing, but lost connection, apologies). If we really wanted to be specific, what we could do would be to trigger a confirmation screen in specific cases, e.g. in descriptions, when users types in >100 chars. But that's something we do not need for MVP. Given the small time left before the end of Delta, dev time should be spent on more pressing issues.

PR #16, opened 2018-04-30
Create service details page

nlaz:
### Description
This PR creates the UI for the service details page. It uses fake data until the endpoint is created.

### Ticket
Related to https://trello.com/c/6VZoL2pT/101-ui-build-service-details-view.

### Screenshots
screen shot 2018-04-30 at 6 19 46 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "REACT_APP_API_URL='https://w6pkliozjh.execute-api.us-east-1.amazonaws.com/prod' react-scripts build",
     "test": "react-scripts test --env=jsdom",
     "eject": "react-scripts eject",
     "lint": "eslint .",
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -15,10 +15,10 @@ export const getLocation = locationId => (dispatch) => {
     .getLocation({
       id: locationId,
     })
-    .then(locationResponse =>
+    .then(data =>
       dispatch({
         type: GET_LOCATION_RESPONSE,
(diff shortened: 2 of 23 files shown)


Rovack:
This looks good to me, just not sure about something: when you say "until the endpoint is created", which endpoint do you mean? If this is referring to the one for getting existing services for a location, IIRC we said you already get that as part of the initial call to `GET /locations/:locationId` (under the "Services" prop).
If you mean something for adding a service, then that should also exist: `POST /services`. It takes the taxonomy ID as well as any number of fields used in the creation of the service (in our case, I imagine you'll always pass at this point `taxonomyId` and `locationId`, I think `name` will be relevant only if the service was added using "Add another" or whatnot, and `description` and `url` will only be added in a later screen).

Is that all related to what selectCategories is supposed to do, and what state locationServices is supposed to store? I wasn't really sure I understood those.

Rovack:
Oh, and one other thing is that there seem to be a lot of changes here, some actually part of other PRs. Is that just a function of what this is branched off of?

nlaz:
@Rovack I didn't realize we could grab service data from the location endpoint. It would be beneficial if we created a `GET` endpoint for `/services/:serviceId` or `/locations/:locationId/services/:serviceId` for the service details page because if someone refreshes that page we shouldn't have to refresh all the location data and parse the services for the required service on this page.

Rovack:
Discussed with @nlaz last night, and decided for now we'd try to use the location endpoint after all, as if someone does refresh the page, we probably do want to fetch all the location data (in case they e.g. try to navigate back to the location info).
If this turns out not to make sense, a GET `/service/:serviceId` route should take under a minute to add.

PR #19, opened 2018-05-03
Create documents required container

nlaz:
### Description
This PR simply sets up the documents required details page with a basic flow between the document fields.

### Ticket
Related to https://trello.com/c/MBg8KJMF/89-ui-build-all-the-screens-for-the-services-forms.

### Screenshots
screen shot 2018-05-02 at 11 23 52 pm


src/app/documents/DocsFormContainer.jsx
```diff
@@ -0,0 +1,60 @@
+import React, { Component } from 'react';
```
Rovack: I'm not sure, but this component looks a lot like the other form container we had (location, service). The only clear differences I can see are the specific routes file and the data fetching action.
Could we not pass those in as props, and reuse 1 component for all these cases?

src/app/documents/DocsFormContainer.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+import Button from '../../components/button';
+import Icon from '../../components/icon';
+
+import DocumentFormRoutes, { DOCUMENT_FIELDS as routes } from './routes';
+import NavBar from '../NavBar';
+import ProgressBar from '../locationInfo/ProgressBar';
+
+const getDocsUrl = (locationId, serviceId) =>
```
Rovack: These URLs are generally encapsulated in the api service.

src/app/service/details/ServiceDetails.jsx
```diff
@@ -1,6 +1,60 @@
...
+
+  render() {
+    const { locationId, serviceId } = this.props.match.params;
+    return (
+      <div className="text-left d-flex flex-column">
+        <NavBar title="Service Details" />
+        <div className="mb-5">
+          <ProgressBar step={1} steps={10} />
+        </div>
+        <ServiceHeader>Check all the Soup Kitchen details</ServiceHeader>
```
Rovack: Since Soup Kitchens aren't special in any way, I assume that part could just be taken from one of the props in the data?

src/app/serviceForm/languages/AddLanguageForm.jsx
```diff
@@ -0,0 +1,77 @@
+import React, { Component } from 'react';
```
Rovack: It looks like this is implementing another search dropdown, like the one @jbeard4 implemented.
Can they not be using the same component?

src/app/serviceForm/openingHours/ServiceOpeningHours.jsx
```diff
@@ -0,0 +1,76 @@
...
+          <div>
+            <p>Is there anything else you would like to add about their opening times?</p>
+            <Input fluid placeholder={inputPlaceholder} />
+          </div>
+        )}
+        {active === 1 && (
+          <div>
+            <p>Select the days and times this service is available</p>
+            <Selector fluid>
+              <Selector.Option active={weekdays[0]} onClick={() => this.onWeekday(0)}>
```
Rovack: It would probably make sense for these weekday numbers to match the ones the API and DB expect (and generally common standards for this), where Monday is 1, and Sunday is 7.

Rovack:
@nlaz @jbeard4 I think it probably does make sense to merge PRs even when there are some open questions (as long as they're not about functionality but about code/architecture), but in those cases I think it's important that we create tech debt cards to track them.

In the case of this specific PR, most of the comments are probably minor enough that they don't require even that, but for something like the search bar (/autocomplete/dropdown/whatever we call it), that's seemingly implemented here even though there's already another one in the map view, I think it is important that we don't forget about it.
If we maintain 2 separate search bars, that's not just a code maintenance issue, but also probably means that our UI won't be fully consistent between the different screens, which is potentially bad UX.
(Of course, this is all assuming what's being implemented here is really another search bar. I'm not sure if it is, but just know that the question about it was never answered.)

jbeard4:
@Rovack Sorry, last night's merge into master was my mix up. I'll work with Niko today on standardizing the autocomplete component.

src/app/service/details/ServiceDetails.jsx
nlaz: Fyi this is updated later

PR #20, opened 2018-05-03
Creating services from user selection

nlaz:
### Description
This PR updates the `ServiceCategories`, `ServiceRecap` and `ServiceDetails` components to create and fetch services based on the taxonomy categories showed on the `ServiceCategories` page. It also fetches existing services and displays those as selected (and disabled) on the categories page.

### Ticket
Related to https://trello.com/c/MBg8KJMF/89-ui-build-all-the-screens-for-the-services-forms

### Screenshots
screen shot 2018-05-03 at 4 34 46 pm


src/app/service/categories/ServiceCategories.jsx
```diff
@@ -21,12 +22,34 @@ const LoadingView = () => (
...
     if (!this.props.taxonomy) {
       this.props.getTaxonomy();
     }
+    if (!this.props.location) {
+      const { locationId } = this.props.match.params;
+      this.props.getLocation(locationId);
+    }
+  }
+
+  componentWillReceiveProps(nextProps, prevState) {
```
Rovack: Shouldn't matter too much here, but I think with React, copying props to unchanging state is often considered an anti-pattern. Was the reason for it here concern about the performance impact of needing to map the location data to categories every time something (unrelated) changes in the state?
If so, I'm not sure if that isn't somewhat premature optimization - we're really just talking about iterating over a small handful of objects.
And even if we were very concerned about the performance of this, there's always the more foundational solutions, like using a memoized selector (e.g. with `reselect`) to map the prop to begin with in mapStateToProps.

src/app/service/details/ServiceDetails.jsx
```diff
@@ -25,21 +28,66 @@ function ServiceHeader({ children }) {
...
+      this.setState({ service });
+    }
+  }
+
   onGoToDocs = () => {
     const { locationId, serviceId } = this.props.match.params;
     this.props.history.push(`${getServiceUrl(locationId, serviceId)}/documents`);
   };
 
+  getServiceById = (location) => {
```
Rovack: Didn't you say the reducer would just take care of this?

src/app/service/categories/ServiceCategories.jsx
nlaz: Was just finding a place to update the current categories. I guess this belongs in the reducer instead.

src/app/service/details/ServiceDetails.jsx
nlaz: I can move this to the reducer 👍

PR #22, opened 2018-05-05
Feature/data freshness labels

jbeard4:
* Add support for saving organization fields in the location info screen
* Add support for freshness labels in the location info screen


src/app/locationInfo/LocationInfo.jsx
```diff
@@ -68,14 +55,26 @@ class LocationInfo extends Component {
...
-          <FieldItem
-            key={route[0]}
-            title={route[2]}
-            linkTo={`${this.props.location.pathname}/${route[0]}`}
-            updatedAt={FAKE_DATA[i]}
-          />
-        ))}
+        {
+          routes.map((route, i) => {
+            const prop1 = route[3];
```
Rovack: The more we add to the routes, the less clear it is to me why it's an array of arrays, requiring all these magic numbers to access props based on index, as opposed to an array of objects, which would allow e.g. just:
```
const { prop1, prop2 } = route;
```

src/app/locationInfo/LocationInfo.jsx
```diff
@@ -68,14 +55,26 @@ class LocationInfo extends Component {
...
+            const prop2 = route[4];
```
Rovack: Feels a bit arbitrary that we assume all fields will always be nested exactly 2 levels deep. An array or dot-delimited string would probably make for a more general representation of paths, and, as you mentioned, allow us to just use a standard "get path" function, from lodash or underscore or whatever.

Rovack: Although, now that I think about it, the last property would always need to be a special case, since it's accessing an array of elements with field_names, as opposed to an object where the field_names are the props.
So, maybe a more generic solution wouldn't be that useful after all.

Perhaps just naming these something a bit more indicative would be nice though, like "metadata section" and "field name" instead of "prop 1" and "prop 2".

src/app/service/categories/ServiceCategories.jsx
```diff
@@ -107,7 +107,7 @@ class ServiceCategories extends Component {
 }
 
 const mapStateToProps = state => ({
-  taxonomy: state.db.taxonomy,
+  taxonomy: state.locations.taxonomy,   //TODO: refactor this so that taxonomy is in its own top-level prop
```
Rovack: I very much agree with this comment, though the one question for me would be: until we do take taxonomy out, is there even a point renaming this to "locations"? As long as taxonomy is under it, calling it "locations" seems more confusing to me than just "db".

Rovack:
Left a few comments about the code, but none of them should have any functional impact I think.
In that sense, all LGTM here.

src/app/locationInfo/LocationInfo.jsx
jbeard4: Fixed in b87dfec18c06b679b1b49a52590e15f67a00c5e5

src/app/locationInfo/LocationInfo.jsx
jbeard4: Renamed prop1 and prop2 to metaDataSection and fieldName in b87dfec18c06b679b1b49a52590e15f67a00c5e5

src/app/service/categories/ServiceCategories.jsx
jbeard4: I'm going to do this refactoring next.

PR #23, opened 2018-05-05
Feature/confirmation

jbeard4:
* Update the freshness label optimistically
* Confirm button saves to the back-end
* Hide up/down buttons when input field has focus

http://g.recordit.co/uVxEpvvVU4.gif

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -42,13 +42,15 @@ export const selectCategories = (locationId, params) => (dispatch) => {
   });
 };
 
-export const updateLocation = (locationId, params) => (dispatch) => {
+export const updateLocation = (locationId, params, metaDataSection, fieldName) => (dispatch) => {
   // optimistically update the data store
   dispatch({
     type: OPTIMISTIC_UPDATE_LOCATION,
     payload: {
       id: locationId,
       params,
+      metaDataSection, 
+      fieldName
     },
   });
   api
@@ -71,7 +73,7 @@ export const updateLocation = (locationId, params) => (dispatch) => {
     });
 };
 
-export const updatePhone = (locationId, phoneId, params) => (dispatch) => {
+export const updatePhone = (locationId, phoneId, params, metaDataSection, fieldName) => (dispatch) => {
   // optimistically update the data store
   dispatch({
     type: OPTIMISTIC_UPDATE_PHONE,
@@ -79,6 +81,8 @@ export const updatePhone = (locationId, phoneId, params) => (dispatch) => {
       locationId,
       phoneId,
       params,
+      metaDataSection, 
+      fieldName
     },
   });
   api
@@ -101,13 +105,15 @@ export const updatePhone = (locationId, phoneId, params) => (dispatch) => {
     });
 };
 
-export const createPhone = (locationId, phoneId, params) => (dispatch) => {
+export const createPhone = (locationId, phoneId, params, metaDataSection, fieldName) => (dispatch) => {
   // optimistically update the data store
   dispatch({
     type: OPTIMISTIC_CREATE_PHONE,
     payload: {
       locationId,
       params,
+      metaDataSection, 
+      fieldName
     },
   });
   api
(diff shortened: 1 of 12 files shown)


src/reducers/locations.js
```diff
@@ -23,29 +24,35 @@ export const locationsReducer = (state = {}, action) => {
...
+            metadata : constructUpdatedMetadata(location, metaDataSection, fieldName, dateString)
           },
         };
       }
       break;
     case OPTIMISTIC_UPDATE_LOCATION:
       if (action.payload) {
+        const { id, params, metaDataSection, fieldName } = action.payload;
+        const location = state[id];
+        console.log('state[id]',state, id, location);
```
nlaz: 🔪 extra console log

src/actions/index.js
```diff
@@ -42,13 +42,15 @@ export const selectCategories = (locationId, params) => (dispatch) => {
...
 
-export const updateLocation = (locationId, params) => (dispatch) => {
+export const updateLocation = (locationId, params, metaDataSection, fieldName) => (dispatch) => {
   // optimistically update the data store
   dispatch({
     type: OPTIMISTIC_UPDATE_LOCATION,
     payload: {
       id: locationId,
       params,
+      metaDataSection,
```
nlaz: What do these new params (`metaDataSection` and `fieldName`) do?

src/reducers/locations.js
```diff
@@ -11,6 +11,7 @@ import {
 } from '../actions';
 
 export const locationsReducer = (state = {}, action) => {
+  const dateString = (new Date()).toISOString();
```
nlaz: What is the `dateString` for?

src/reducers/locations.js
```diff
@@ -101,15 +108,34 @@ export const locationsReducer = (state = {}, action) => {
...
       Phones: newPhones,
+      metadata: constructUpdatedMetadata(location, metaDataSection, fieldName, dateString)
     },
   };
 }
 
+function constructUpdatedMetadata(location, metaDataSection, fieldName, dateString){
+  const metadata = location.metadata;
+  const subFields = metadata[metaDataSection];
+  const fieldIndex = subFields.findIndex( field => field.field_name === fieldName );
```
nlaz: Suggestion:
```js
const newField = { field_name: fieldName, last_action_date: dateString };
const newSubFields = subFields.map(field => field.field_name === fieldName ? newField : field );
return { ...metadata, [metaDataSection]: newSubFields  };
```

src/app/locationForm/LocationForm.jsx
```diff
@@ -73,7 +96,13 @@ class LocationForm extends Component {
           <div className="container">
             <div className="row px-4">{this.routeComponents}</div>
           </div>
-          <div className="position-absolute" style={{ right: 0, bottom: 12 }}>
+          <div 
+            className="position-absolute" 
+            style={{ 
+             right: 0,
+             bottom: 12,
+             display: this.state && this.state.inputFocused ? 'none' : 'block'
```
nlaz: I didn't realize the arrows should be hidden on input focus. Does that apply to every page?

src/actions/index.js
jbeard4: These are used to look up the last_updated_date field on the location metadata object. We pass them through to the reducer so that the reducer is also able to look up the metadata field in the location object in the store, and optimistically update the last_updated_date field in the store, so that the we can update the freshness labels in the UI after the field as been updated/confirmed.

src/app/locationForm/LocationForm.jsx
jbeard4: This is to work around a behavior in Android, where the open keyboard reduces the page size. Because the buttons are absolutely positioned, they stick to the bottom of the resized page, which looks like this:

</img>

See row 16 on the master bug list.

src/reducers/locations.js
jbeard4: Looks good.

src/reducers/locations.js
jbeard4: This is to update the metadata last_updated_date.

PR #24, opened 2018-05-05
Service form editing fields

nlaz:


jbeard4:
This is a big PR, which involved a lot of refactoring, so I wanted to describe the steps I have been taking for doing ad-hoc testing and evaluating the UI:

1. test that data is updated in the store after successful save, and before a page refresh:
```
Click form A
Click "No, Let's Edit it"
Edit value.
Click Save
(Page will automatically advance)
Click Up Arrow. 
Verify the data matches the data that was entered.
```

2. test that freshness field has updated after save:
```
Click Back button in NavBar.
Verify that data freshness label says "A few seconds ago"
```

3. test that updated data value is shown in the UI after page refresh:
```
Refresh page.
Click into the form
Verify that the data matches the data that was entered.
Click Back button in NavBar.

4. test for verify form success:
```
Click a different form
Click Yes in "Is this information Correct" field
Click Back button in NavBar.
Verify that the data freshness label says "A few seconds ago.
```

<img src="http://g.recordit.co/AToqo4Ym4R.gif"></img>

The current issues I am seeing are:

- [ ] Regression in Locations page. Test 3 fails: the freshness labels no longer update without page refresh.
- [ ] In Services page, `Service description` form fails tests 1, 2, and 4. Passes 3.
- [ ] In Services page and Documents page, all other forms fail tests 1, 2, 3, 4
- [ ] Throws an error when clicking OK on last screen of Services flow

jbeard4:
Couple issues:

"Who does it serve?" does not persist across page refresh:

</img>

jbeard4:
"Ages served" fails tests 1 and 3 (but passes 2).


</img>

jbeard4:
Thanks screen needs to match the copy in the specification. Also, should blur the background, like it currently does in the Thanks screen in the Location flow.

screen shot 2018-05-08 at 8 57 18 pm
screen shot 2018-05-08 at 8 58 20 pm

PR #30, opened 2018-05-10
Ui fixes

jbeard4:


src/app/service/recap/ServiceRecap.jsx
```diff
@@ -70,7 +70,6 @@ class ServicesRecap extends Component {
       return <LoadingView />;
     }
 
-    console.log('props', this.props);
     // const showThanks = true;
```
nlaz: Ah oops! This is my fault. Thanks for cleaning this up.

PR #112, opened 2018-08-18
Feature/show network error

### LINKED ISSUE #84: How do we show the user error messages, especially network errors?
Reporter: JB

jbeard4:
Fixes #84 

This feature is not yet complete, as we need to apply it to services pages, and other call sites that make network requests. If you like the technique presented here, I will update this PR to apply it to other pages.

Should be merged after branch feature/who_does_it_serve2


src/reducers/locations.js
```diff
@@ -40,6 +42,20 @@ function constructNewStateWithUpdatedPhones(state, action, newPhones, location,
   };
 }
 
+export const locationErrorsReducer = (state = {}, action) => {
```
Rovack: Any reason to keep 2 high-level reducers in the same file?

src/actions/index.js
```diff
@@ -1,7 +1,9 @@
 /* eslint-disable no-console */
 import * as api from '../services/api';
 
+export const GET_LOCATION_REQUEST = 'GET_LOCATION_REQUEST';
 export const GET_LOCATION_RESPONSE = 'GET_LOCATION_RESPONSE';
+export const GET_LOCATION_RESPONSE_ERROR = 'GET_LOCATION_RESPONSE_ERROR';
```
Rovack: I think usually this would just be "GET_LOCATION_ERROR" (and the other 2 would be "GET_LOCATION_REQUEST" and "GET_LOCATION_SUCCESS").

src/reducers/locations.js
Rovack: (I mean, I think keeping it in the same file could make sense, but then also probably keeping it in the same reducer, as just an "error" property and a "data" or whatnot property in the same object, to avoid the 2 switches listening to effectively the same actions and managing closely related state.)

src/components/form/ErrorLabel.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import './ErrorLabel.css'
+
+const ErrorLabel = ({ errorMessage }) => (
+  <div className="FormEdit">
+    <p>
+      <i className="fas fa-exclamation-triangle" aria-hidden="true" />&nbsp;
+      <span>{errorMessage}</span>
+    </p>
+    <p onClick={() => {window.location.reload()}}> Click here to reload the page </p>
```
Rovack: Creating a new anonymous function here every render seems unnecessary. Why not just `onClick={window.location.reload}`?

src/components/form/ErrorLabel.css
```diff
@@ -0,0 +1,14 @@
+.FormEdit {
```
Rovack: Shouldn't this use a ".ErrorLabel" class?
Is it that you wanted it to share some CSS with the actual form?

src/app/locationInfo/LocationInfo.jsx
```diff
@@ -57,7 +59,11 @@ class LocationInfo extends Component {
   };
 
   render() {
-    const { locationData } = this.props;
+    const { locationData, locationError } = this.props;
+
+    if(locationError){
+      return <ErrorLabel errorMessage={this.props.locationError} />;
```
Rovack: So I guess this will only work for the location form, not service/documents/comments/etc?
Would be nice if we had some "base layout" type thing that handled this, or some way for the router to always recognize there was an error and show it accordingly... but maybe that's difficult.

src/reducers/locations.js
jbeard4: I don't want to do the refactoring that would need to update all the call sites that reference state.location, so that state.location maps `locationId -> {locationData, errorData}`, instead of `locationId -> locationData`. I am a bit concerned about breaking some existing behavior. I would therefore rather break errors out into a separate top-level state.

src/actions/index.js
jbeard4: Sounds good. Will fix.

src/components/form/ErrorLabel.jsx
jbeard4: Try this:

```javascript
f = window.location.reload
f()
```
You will get error:

```
VM54:1 Uncaught TypeError: Illegal invocation
    at <anonymous>:1:1
(anonymous) @ VM54:1
```

So, basically, window.location.reload must be called as a method of the location object, and bound to location. It would be sufficient to break the anonymous function out into a named function so that a new anonymous function is not declared on each render.

src/components/form/ErrorLabel.css
jbeard4: ah, that's a typo. Will fix.

src/app/locationInfo/LocationInfo.jsx
jbeard4: Yes, as I mentioned in my comment above, this will currently only work for locations. Once we decide whether we like the pattern outlined in this PR, I can go through and apply the same pattern to service and documents. Eventually, we will want to generalize these patterns into some reusable action/reducer code.

src/actions/index.js
jbeard4: Fixed in 86e4e86

src/components/form/ErrorLabel.jsx
jbeard4: Fixed in 91a8dbb

src/components/form/ErrorLabel.css
jbeard4: Fixed in eb37e32

src/reducers/locations.js
jbeard4: Actually, I took another look, and refactoring does not seem too difficult. I'll go ahead and do this.

jbeard4: I made a first pass at refactoring, and while all of the view code is easily updatable, the reducer code that handles this is non-trivial. Because we do not yet have automated tests, it is challenging to do a refactoring of the state without introducing bugs. So, I am going to go back to my original suggestion, which is that we do not change the locations data structure at this time, and instead introduce a new top-level state for errors.

src/components/form/ErrorLabel.jsx
```diff
@@ -1,13 +1,17 @@
...
+}
+
 const ErrorLabel = ({ errorMessage }) => (
   <div className="FormEdit">
     <p>
       <i className="fas fa-exclamation-triangle" aria-hidden="true" />&nbsp;
       <span>{errorMessage}</span>
     </p>
-    <p onClick={() => {window.location.reload()}}> Click here to reload the page </p>
+    <p onClick={reload}> Click here to reload the page </p>
```
Rovack: Actually my comment was just that `window.location.reload` could be passed directly, no need to wrap in another function (anonymous or otherwise).
Or, was there some reason that doesn't work?

jbeard4:
@Rovack I think this is ready for review. I tested all routes. 

Going forward, it could be extended to handle more network errors in `/services/api.js`, such as, for example, posting comments.

src/reducers/locations.js
Rovack: Yup, makes sense to keep it as separate high-level state then.
Back to the original question, though: any reason why the 2 reducers should be in the same file? (As opposed to a src/reducers/errors or src/reducers/locationErrors or something?)

src/reducers/locations.js
```diff
@@ -40,6 +42,18 @@ function constructNewStateWithUpdatedPhones(state, action, newPhones, location,
...
```
Rovack: Btw, is it only locations that can break a page by failing to load, or are there cases where other errors actions or pieces of state could do the same?

src/app/locationInfo/LocationInfo.jsx
jbeard4: > Right, I think the last part is more what I was referring to.
If we had a "base layout" component that's used for all screens, we wouldn't have to go into all the different pages and change them, whereas the current way seems to introduce a bit of duplication.

Yes, it would definitely be good to start to consolidate this code so that we don't have to reduplicate the "loading screen"/"error screen" logic all over the place.

> I suppose in terms of the Redux side of things, it would require a reducer that listens to error actions in general (either through knowing all the specific ones, or looking for something standardized like an "error" property in the payload), and just puts any errors in a single piece of state. This kind of error reducer is in fact pretty standard, and probably requires less code than the multiple different error reducers. If it's really that simple, and avoids duplication, might be good to just do it that way.

I already do this a bit in the Form component, and the sites that use. You can see this here: 
https://github.com/streetlives/streetlives-web/pull/112/files#diff-a70d9e4934a427002e689cb0166f5d22R61

src/app/comments/newCommentForm/NewCommentForm.jsx
```diff
@@ -1,8 +1,10 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { postComment, getLocation } from '../../../actions';
-import { selectLocationData, selectIsPostingComment } from '../../../reducers';
+import { selectIsPostingComment } from '../../../reducers';
+import { getLocationError, getLocation as selectLocationData } from '../../../selectors/location';
```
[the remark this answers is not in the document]
jbeard4: There was already a `getLocation` selector in `selectors/location`. I aliased it to `selectLocationData` in some modules, where it was necessary to do so because of a naming conflict with `actions/getLocation`, but it might be better to change it at the source, inside of `selectors/location`.

src/app/mapView/MapView.jsx
```diff
@@ -142,7 +142,7 @@ export default class MapView extends Component {
             {
               this.state.suggestions && <Dropdown options={
                 this.state.suggestions.map(organization => ({
-                  key: organization.id,
+                  id: organization.id,
```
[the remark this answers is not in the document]
jbeard4: I fixed this because the Dropdown was looking for an id property, and was complaining because no id property was specified. You can see this in its proptypes:

```
Dropdown.propTypes = {
  options: PropTypes.arrayOf(PropTypes.shape({
    id: PropTypes.string.isRequired,
    label: PropTypes.string.isRequired,
    onClick: PropTypes.func.isRequired,
  })).isRequired,
```

src/reducers/locations.js
jbeard4: I need to look at this more closely. I would start by adding error handlers to other functions in `services/api.js`, adding logic to the reducer to ensure that they propagate the error into the state, and then looking at the components that use the state where a response is expected. I can create a separate task to track this.

src/app/comments/newCommentForm/NewCommentForm.jsx
```diff
@@ -48,7 +50,11 @@ class NewCommentForm extends Component {
   }
 
   render() {
-    const { locationData, isPostingComment } = this.props;
+    const { locationData, isPostingComment, locationError } = this.props;
+
+    if(locationError){
```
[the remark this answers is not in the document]
jbeard4: I think we have an open issue for this https://github.com/streetlives/streetlives-web/issues/113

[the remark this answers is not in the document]
Rovack: Just checked: master has *42* errors, and basically of only 2 types: either jsx-a11y or no-extraneous-dependencies (those 2 types account for 40 of the 42). They affect *16* files in total.

In this branch, there are currently *243* errors, affecting *49* files.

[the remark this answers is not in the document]
Rovack: So, while I tend to agree that 42 is also a lot, the fact there's currently practically only 2 types means they could be solved more or less at once, I hope.
Introducing 200 new errors of various kinds, however, I think is a bigger problem.

Therefore, I stand by my opinion that it would be best to first fix the linter (whether the existing 42 errors are also fixed or not), and only then merge a PR that touches so many files.

src/app/comments/newCommentForm/NewCommentForm.jsx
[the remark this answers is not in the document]
Rovack: Ah, I see.
Then yeah, it would definitely be good to rename getLocation at the source. But either way, if we want the convention to be "select" going forward, we shouldn't be introducing new selectors with the "get" convention. (I.e. getLocationError.)

src/app/locationInfo/LocationInfo.jsx
Rovack: Right, the Form example is great.

I was thinking that this might be a good chance to do that consolidation, seeing as you already need to touch all these sites anyway (so we could just as easily make them all use that core layout instead of making them use ErrorLabel specifically).

But since you've already written all this, if you don't feel like changing it, I suppose we can do it some other time.

src/reducers/locations.js
Rovack: Cool. I think the convention of having a reducer per file is generally a good one.
(If only for consistency.)

src/app/comments/newCommentForm/NewCommentForm.jsx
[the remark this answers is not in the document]
jbeard4: This is fixed in 3c253d8599bfbf6c2e803ce05fe0a7ec1b1c05d8

src/app/locationInfo/LocationInfo.jsx
jbeard4: I created an issue for this: https://github.com/streetlives/streetlives-web/issues/124

src/reducers/locations.js
jbeard4: This is done in 160bb60c6149c94ec0dcc2dba5dcb96dec0ac629

package.json
```diff
@@ -44,7 +44,7 @@
     "build": "REACT_APP_API_URL='https://w6pkliozjh.execute-api.us-east-1.amazonaws.com/Stage' react-scripts build",
     "test": "react-scripts test --env=jsdom",
     "eject": "react-scripts eject",
-    "lint": "eslint . --ext js,jsx",
+    "lint": "eslint src/ --ext js,jsx --ignore-pattern \"*.story.jsx\"",
```
Rovack: I think stories should generally follow our eslint rules too.
If actually fixing them right now would be a bit much, I think that's fine. You could suppress the few specific errors we have in those files, or even leave them for now, just as long as we still notice any new kinds of lint errors we add to the stories.

.eslintrc.js
```diff
@@ -24,5 +24,6 @@ module.exports = {
       },
     ],
     'jsx-a11y/label-has-for': [2, { required: 'id' }],
+    'no-param-reassign': ["error", { "props": false }],
```
Rovack: What was the reason for this? In what case does it make sense to reassign a prop?

package.json
jbeard4: I disabled checking stories because I got an error on lint:

`'@storybook/react' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies`

This error seems incorrect - @storybook/react dependency should definitely be in devDependencies. So I decided to disable checking of stories. Let me know if you would prefer a different approach. Thanks.

.eslintrc.js
jbeard4: Fixed in fe1c08403352a7e5c18fa8bb85b47b6e350bd476

package.json
Rovack: Yeah, when I said "suppress", I meant just using `// eslint-disable-next-line import/no-extraneous-dependencies`.

Although now I did a quick search, and saw this: https://github.com/storybooks/storybook/issues/1992.
So, looks like the best option would be replacing this line with:
```
'import/no-extraneous-dependencies': [2, { devDependencies: ['**/*.story.jsx'] }],
```

.eslintrc.js
Rovack: Ah great, the new method is way better.
Definitely good to avoid bypassing/mutating the component state.

PR #128, opened 2018-10-11
Fixes #126: Flash of unrendered login screen

### LINKED ISSUE #126: Flash of unrendered login screen on navigate on production site https://streetlives.nyc/

jbeard4:


Rovack:
Not sure why autocomplete is bad for phone numbers, but otherwise looks good.

PR #129, opened 2018-12-08
Feature/comment speech to text

Rovack:
Updating the comment entry page to support speech-to-text.

Note: I'm not sure yet whether we actually want this feature and re-design or not, so maybe we shouldn't merge this without further discussion. Just wanted to have this in a PR so we don't forget about it.


Rovack:
@jbeard4 Looks like at least for now we're still giving this speech-to-text thing a shot, so this branch should probably go into master.
Are there any comments about the code that I should fix before merging?

Rovack:
I'll take the silence as tacit agreement. :P

PR #147, opened 2020-05-08
Feature/error reporting for locations

cadecameron:
Hey @Rovack ,

Take a look at the following pull request, which sets-up the frontend error reporting workflow under the `find` route.

Note: I haven't added a button to navigate to the begining of the error reporting flow yet. I'll do that small step in a separate PR.


src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+    this.onErrorReportTextChanged = this.onErrorReportTextChanged.bind(this);
+    this.onErrorReportServicesChanged = this.onErrorReportServicesChanged.bind(this);
+    this.onErrorReportGeneralLocationErrorChanged = this.onErrorReportGeneralLocationErrorChanged.bind(this);
+    this.onErrorReportFinished = this.onErrorReportFinished.bind(this);
+    this.onErrorReportSubmitted = this.onErrorReportSubmitted.bind(this);
+  }
+
+  componentDidMount() {
+    if (!this.props.locationData) {
+      this.props.getLocation(this.props.match.params.locationId);
```
Rovack, line 30: To simplify the code of the component itself, I'd move the knowledge of how to get locationId out of here and into where getLocation is defined below.
This would also make it easier if you ultimately want to split the more "container-y" parts of this file into an actual separate container component.

See e.g. LocationDetailsContainer.js vs LocationDetails.jsx.

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+      };
+    });
+  }
+
+  onErrorReportGeneralLocationErrorChanged(value) {
+    this.setState({ errorReportGeneralLocationError: value });
+  }
+
+  onErrorReportFinished() {
+    this.setState({ isErrorReportFinished: true });
```
Rovack, line 60: This naming is confusing, since it's not the error report which is finished, but only the step where the user indicates where the errors are.
The state and function should be renamed to reflect this.

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  onErrorReportSubmitted() {
+    console.log('onErrorReportSubmitted triggered.');
```
Rovack, line 64: Same comment as elsewhere, that console logs should be removed before merging (don't want them showing up in actual users' browsers).

src/app/errorReports/newErrorReportForm/ErrorReportText.jsx
```diff
@@ -0,0 +1,72 @@
...
+              className="border" // flex-1 py-2 pl-2 pr-2
+            />
+          </div>
+          <div className="mx-4 mb-1">
+            <Button
+              onClick={this.props.onSubmit}
+              primary
+              fluid
+            >
+              Submit
```
Rovack, line 48: What about the Skip state?

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+
+  render() {
+    const { isPostingErrorReport, locationError } = this.props;
+
+    if (locationError) {
+      return <ErrorLabel errorMessage={locationError} />;
+    }
+
+    // Comments component checks for (!locationData || isPostingErrorReport)
+    // Not sure what the !locationData check was doing specifically
```
Rovack, line 83: It's for while the location data is being fetched, to make sure a loading indicator is shown.
Try refreshing the page (preferably while using throttling in Chrome's dev tools), and see what happens while it's loading.

I think the only reason you're not seeing actual crashes without this check is that right now, location data is coming to the child components not from here, but rather from the `withErrorReportsForm`, which does have a `if (!locationData) {` check.

src/app/errorReports/thanks/Thanks.jsx
```diff
@@ -0,0 +1,44 @@
...
+import withErrorReportsForm from '../withErrorReportsForm';
+
+class Thanks extends Component {
+  constructor(props) {
+    super(props);
+    this.goToViewLocation = this.goToViewLocation.bind(this);
+  }
+
+  goToViewLocation() {
+    this.props.history.push(`/find/location/${this.props.match.params.locationId}/view`);
```
Rovack, line 13: The route is actually just `/find/location/${this.props.match.params.locationId}`, no "/view".
I don't think it actually works with "/view" at the end. If anything, it likely takes the user back to the map, which is what happens in general with URLs that end in unexpected patterns.

src/app/errorReports/thanks/Thanks.jsx
```diff
@@ -0,0 +1,44 @@
+import React, { Component } from 'react';
```
Rovack, line 1: IIRC in the designs this was actually a modal on top of the LocationDetails/NewErrorReportForm modal, not its own full page.

src/services/api.js
```diff
@@ -290,3 +290,42 @@ export const removeComment = comment =>
...
+      url: `${config.baseApi}/errorreports`,
+      method: 'get',
+      params: {
+        locationId,
+      },
+    })
+    .then(result => result.data);
+
+export const postErrorReport = ({ locationId, errorReport }) =>
+  getAuthToken() // Remove authorization requirement for posting new error reports?
```
Rovack, line 307: Yes, definitely shouldn't require auth.
I believe the backend already doesn't require it.

src/services/api.js
```diff
@@ -290,3 +290,42 @@ export const removeComment = comment =>
         Authorization: idJwtToken,
       },
     }));
+
+/* ------ Error Report API functions -------- */
+export const getErrorReports = ({ locationId }) =>
```
Rovack, line 295: Here, too, no need to keep the dead code around.
Only need postErrorReport.

src/app/errorReports/withErrorReportsForm.jsx
```diff
@@ -0,0 +1,110 @@
+import React, { Component } from 'react';
```
Rovack, line 1: I imagine this was taken from comments, but I think for error reports this Higher-Order Component is redundant and could just be removed.

I mean, the data fetching and all that is handled by the parent NewErrorReportForm component anyway (currently it fetches locationData but doesn't actually pass it to the components, but that would just take 1 or 2 lines to fix), and any high-level layout elements needed (i.e. the Modal component, I guess) could be part of NewErrorReportForm too.

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
```diff
@@ -0,0 +1,130 @@
+import React, { Component } from 'react';
```
Rovack, line 1: "NewErrorReportForm" sounds like a reasonable name for the component, but for the directory, doesn't "errorReport" convey basically the same meaning as "newErrorReportForm"?
And arguably, the Thanks component is part of the form as well (especially in the designs, where it's just a modal on top of it), so the additional subdirectory might not really be needed.

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+    // Not sure what the !locationData check was doing specifically
+    if (isPostingErrorReport) {
+      return (
+        <LoadingLabel>
+          <span>Adding error report...</span>
+        </LoadingLabel>
+      );
+    }
+
+    if (!this.state.isErrorReportFinished) {
```
Rovack, line 92: The fact that the transition between selecting checkboxes and entering text is handled by this component's internal state means that if the user hits "Back" while on the text section, they'll go back to location details, not to the checkboxes.
They'll lose whatever checkboxes they selected, and there's certainly a chance they'll feel frustrated and just never submit the report.

Fixing this, however, would require some changes:
* The checkboxes and the text sections would need to be 2 separate routes under the Router
* The state gathered in the checkboxes section would need to be stored in Redux state
* The text section would need to check if checkboxes state exists, and if not, redirect back to that section (this is in case e.g. the user refreshes the page while on the text section, as Redux state isn't currently persisted across refresh)

This probably isn't critical. It can create a pretty bad user experience, but only if the user actually hits Back after selecting the checkboxes, while likely won't be too common.
If you think you can make these changes pretty easily, I'd say go for it. Otherwise, we can probably do without.

Either way, I would keep it for last.

Rovack:
Oh, and to your question of what appears in the "services" checkboxes, I noted near the relevant place in the code that what you did is exactly right. Asked Di, and it should just be the service names; no need for fancy grouping and such.

cadecameron:
Wow - thanks for taking a look at this (admittedly large) PR @Rovack.
Your comments will be esp. helpful when I'm making implementing these changes - there's some great guidance and ideas/concepts in them, and I appreciate the time you spent writing them! 👍

src/app/App.jsx
```diff
@@ -54,6 +55,7 @@ function App() {
                 />
               ))}
               <Route path="/comments" component={withTracker(CommentsRouter)} />
+              <Route path="/errorreports" component={withTracker(ErrorReportsRouter)} />
```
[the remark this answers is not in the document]
cadecameron: Yeah, after starting to test the components on screen, I started thinking about this concept as well, and how it was potentially different to the comment feature.

I'll take a look at the easiest way to implement this.

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
cadecameron: Yeah, I thought the same, and considered renaming it, however I was trying to remain consistent with the `comments` naming.

Perhaps something like `isErrorReportFinalScreen` would be more verbose.

src/app/errorReports/newErrorReportForm/ErrorReportText.jsx
```diff
@@ -0,0 +1,72 @@
...
+        </div>
+        <div>
+          <div className="mx-4 mb-1">
+            <TextArea
+              value={this.props.errorReportText}
+              onChange={this.onTextChange}
+              minRows={7}
+              autoFocus
+              fluid
+              className="border" // flex-1 py-2 pl-2 pr-2
```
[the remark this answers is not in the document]
cadecameron: It was a backup. I'll remove it.

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
Rovack: Sure, that sounds good.
Following that thinking, you could be even more explicit, with something like "isOnErrorContentScreen".
Anything that suggests only the selection of services is done but the report isn't quite done yet.

src/app/errorReports/newErrorReportForm/ErrorReportServices.jsx
```diff
@@ -0,0 +1,107 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import Header from '../../../components/header';
+import Button from '../../../components/button';
+import Checkbox from '../../../components/checkbox';
+import withErrorReportsForm from '../withErrorReportsForm';
+
+/*
```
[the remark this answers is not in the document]
cadecameron: Yeah, was just the props format for the Checkbox component I was using for reference.
Removed.

src/app/errorReports/newErrorReportForm/ErrorReportServices.jsx
```diff
@@ -0,0 +1,107 @@
...
+              checked={generalLocationError}
+              onChange={this.onGeneralLocationChange}
+            />
+            <div />
+            {locationData.Services.map(service => (
+              <div key={service.id}>
+                <Checkbox
+                  key={service.id}
+                  name={service.id}
+                  label={`${service.name} service`} //  || service.Taxonomies[0].name || service.Taxonomies[0].parent_name
```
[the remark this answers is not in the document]
cadecameron: Removed the " service".

src/app/errorReports/newErrorReportForm/ErrorReportServices.jsx
```diff
@@ -0,0 +1,107 @@
...
+          </div>
+          <div>
+            <Checkbox
+              key="generalLocationError"
+              name="generalLocationError"
+              label="Information about the location"
+              checked={generalLocationError}
+              onChange={this.onGeneralLocationChange}
+            />
+            <div />
```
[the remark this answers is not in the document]
cadecameron: It was pushing the location checkboxes onto a new line.
Removed for the moment.

src/app/errorReports/newErrorReportForm/ErrorReportServices.jsx
```diff
@@ -0,0 +1,107 @@
+import React, { Component } from 'react';
```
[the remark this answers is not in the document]
cadecameron: Changed to `ErrorReportInformationSelect`

src/app/errorReports/newErrorReportForm/ErrorReportServices.jsx
```diff
@@ -0,0 +1,107 @@
...
+    );
+  }
+}
+
+ErrorReportServices.propTypes = {
+  onServiceChange: PropTypes.func.isRequired,
+  onGeneralLocationChange: PropTypes.func.isRequired,
+  onSubmit: PropTypes.func.isRequired,
+  generalLocationError: PropTypes.bool.isRequired,
+  errorReportServices: PropTypes.instanceOf(Array),
```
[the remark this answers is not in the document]
cadecameron: Added `locationData` check, and no, eslint doesn't raise an error for the `Array` reference. 🤔
(At least, not when I run `npm run lint`)

src/app/errorReports/newErrorReportForm/ErrorReportText.jsx
cadecameron: Di thought that `Submit` might be confusing for users, and instead suggested `Done` (static value) for the button. I've updated to this wording.
In addition, there's no additional step to 'skip' to - this is the final screen before the error report is submitted.

src/app/errorReports/thanks/Thanks.jsx
cadecameron: Oh, good catch! The `/view` was taking me back to the location information page, since the Route was matching the start of the url (and `exact` wasn't being used), but it wasn't necessary.

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
cadecameron: Yeah, I was replicating the comments directory structure with this.
I can flatten the `errorReports` directory, which should make things simpler when I integrate into the `find` flow.

src/app/errorReports/newErrorReportForm/ErrorReportServices.jsx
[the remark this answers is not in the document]
Rovack: Ahh, I see why: https://github.com/yannickcr/eslint-plugin-react/issues/1719.
If you were to go with the standard PropTypes usage for arrays, i.e. `PropTypes.array`, you _would_ get a warning, but apparently eslint isn't clever enough to catch usages of instanceOf.

In general, it sounds like you should avoid `PropTypes.instanceOf`. The only exception would be if there's an instance of an actual custom class that's expected to be passed in as a prop, but there shouldn't be many/any cases of that.

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+    this.onErrorReportSubmitted = this.onErrorReportSubmitted.bind(this);
+  }
+
+  componentDidMount() {
+    if (!this.props.locationData) {
+      this.props.getLocation(this.props.match.params.locationId);
+    }
+  }
+
+  componentDidUpdate(prevProps) {
```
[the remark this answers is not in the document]
cadecameron: Would one of these be the preferred way to get the thunk to complete the routing?
https://stackoverflow.com/questions/35706835/react-router-redirect-after-action-redux

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
cadecameron: Another option (while not exactly fixing the user hits "Back") would be to include a `Back` button between the `Next`/`Done` button and the `Cancel` button. This would enable the user to at least go back and modify their previous selection, without having to hook up Redux/new routes, etc. (though that _is_ probably the optimal end-goal eventually).

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+    if (prevProps.isPostingErrorReport && !this.props.isPostingErrorReport) {
```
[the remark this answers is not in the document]
cadecameron: Removed wait when posting new error report.

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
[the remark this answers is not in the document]
cadecameron: I navigated directly from the components, however the routing of these components is something I'd like to have a crack at refactoring in the future.

src/app/errorReports/newErrorReportForm/ErrorReportServices.jsx
```diff
@@ -0,0 +1,107 @@
...
+          >
+            Cancel
+          </Button>
+        </div>
+      </div>
+    );
+  }
+}
+
+ErrorReportServices.propTypes = {
```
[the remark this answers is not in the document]
cadecameron: We can discuss refactoring this in a future PR -- I might want to dive into this a bit more with you to really understand this.

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+import { postErrorReport, getLocation } from '../../../actions';
+import { selectIsPostingErrorReport } from '../../../reducers';
+import { selectLocationError, selectLocationData } from '../../../selectors/location';
+import LoadingLabel from '../../../components/form/LoadingLabel';
+import ErrorLabel from '../../../components/form/ErrorLabel';
+import ErrorReportText from './ErrorReportText';
+import ErrorReportServices from './ErrorReportServices';
+
+class NewErrorReportForm extends Component {
+  constructor(props) {
```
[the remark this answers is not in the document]
cadecameron: I like this format better. I'll try to adopt it going forward.

src/app/errorReports/newErrorReportForm/NewErrorReportForm.jsx
cadecameron: Currently the user 'mid-way' check isn't being performed (i.e. a user could navigate directly to the `/text` URL endpoint and continue the workflow from there).
This might be an opportunity to refactor in the future.

src/app/find/locationDetails/locationDetails.css
```diff
@@ -18,24 +18,37 @@
 .categoryCard {
```
Rovack, line 18: How come any styles for the existing pages had to change (except where it's related to adding the button at the bottom)?
Or were some of these changes merge issues or something?

PR #148, opened 2020-05-08
Feature/collapsible categorycard

Vikariusu:
feat: make CategoryCard collapsible, reorder searched category to the top, destructure CategoryCard and add ShelterService as a new component

Link to Asana with a demo: https://app.asana.com/0/1168502694533225/1174572971197987

### DIFF
--- a/src/app/find/locationDetails/CategoryCard.jsx
+++ b/src/app/find/locationDetails/CategoryCard.jsx
@@ -1,40 +1,36 @@
-import React, { Component } from 'react';
+import React from 'react';
 import { getCategoryIcon } from '../../../services/iconography';
 import Icon from '../../../components/icon';
 import './locationDetails.css';
-import ShelterService from './ShelterService';
+import ServiceSection from './ServiceSection';
 
 // this card contains a category (e.g. Food) and displays
 // all services that a shelter provides within that category
-class CategoryCard extends Component {
-  state = {
-    collapsed: false,
-  }
-
-  handleCollapse = () => {
-    this.setState({ collapsed: !this.state.collapsed });
-  }
-
-  render() {
-    const { category, services, isSearchCategory } = this.props;
-
-    return (
-      <div className="categoryCard">
-        <div className="serviceCategoryHeadersContainer">
+function CategoryCard({ category, services, isInitiallyExpanded }) {
+  return (
+    <div className="categoryCard">
+      <div className="serviceCategoryHeadersContainer">
+        <span className="categoryIcon">
           <Icon
             name={getCategoryIcon(category)}
             size="lg"
           />
-          <div className="serviceCategoryHeaders">{category}</div>
-        </div>
-        <div>
-          {services
-            .sort((service1, service2) => service2.lastUpdate - service1.lastUpdate)
-            .map(service => <ShelterService service={service} selectedCategory={false} key={service.id} showOnMount={isSearchCategory} />)}
-        </div>
+        </span>
+        <div className="serviceCategoryHeaders">{category}</div>
       </div>
-    );
-  }
+      <div>
+        {services
+          .sort((service1, service2) => service2.lastUpdate - service1.lastUpdate)
+          .map(service => (
+            <ServiceSection
+              service={service}
+              key={service.id}
+              showOnMount={isInitiallyExpanded}
+            />
+          ))}
+      </div>
+    </div>
+  );
 }
 
 export default CategoryCard;
--- a/src/app/find/locationDetails/CategoryCard.jsx
+++ b/src/app/find/locationDetails/CategoryCard.jsx
@@ -1,169 +1,40 @@
-import React, { Fragment, Component } from 'react';
-import moment from 'moment';
-import { DAYS, OCCASIONS } from '../../../Constants';
+import React, { Component } from 'react';
 import { getCategoryIcon } from '../../../services/iconography';
 import Icon from '../../../components/icon';
-import PhoneLink from '../../../components/phoneLink';
-import InfoItem from './InfoItem';
-import ServiceRestrictions from './ServiceRestrictions';
-import ServiceOfferings from './ServiceOfferings';
 import './locationDetails.css';
+import ShelterService from './ShelterService';
 
-const renderSchedule = (schedule) => {
-  if (schedule.length === 7 &&
-    schedule.every(day => day.opens_at === '00:00:00' && day.closes_at === '23:59:00')) {
-    return 'Open 24/7';
-  }
-
-  const dayNumberToName = weekday => DAYS[weekday - 1];
-
-  const formatHour = time => moment(time, 'HH:mm:ss').format('LT').replace(':00 ', ' ');
-  const formatHours = (opens, closes) => `${formatHour(opens)} to ${formatHour(closes)}`;
-
-  const formatRange = ({ start, end }) => {
-    if (end === start) {
-      return dayNumberToName(start);
-    }
-    if (end === start + 1) {
-      return `${dayNumberToName(start)} & ${dayNumberToName(end)}`;
-    }
-    return `${dayNumberToName(start)} to ${dayNumberToName(end)}`;
-  };
-
-  const orderedSchedule = schedule.sort(({ weekday: day1 }, { weekday: day2 }) => day1 - day2);
-
-  const daysGroupedByHours = orderedSchedule.reduce((grouped, day) => {
-    const hoursString = formatHours(day.opens_at, day.closes_at);
-    return { ...grouped, [hoursString]: [...(grouped[hoursString] || []), day.weekday] };
-  }, {});
-
-  const groupStrings = Object.keys(daysGroupedByHours).map((hoursString) => {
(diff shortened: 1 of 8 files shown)


src/app/find/locationDetails/CategoryCard.jsx
```diff
@@ -1,150 +1,40 @@
...
-        {isClosed && <span className="coronavirusInfo"> (suspended)</span>}
-      </div>
-
-      {!!(coronavirusInfo && coronavirusInfo.length) && (
-        <InfoItem coronavirus icon="exclamation-triangle">
-          {coronavirusInfo[0].information}
-        </InfoItem>
-      )}
+  render() {
+    const { category, services, isSearchCategory } = this.props;
```
Rovack, line 19: Just a nitpick, but there's something to be said for encapsulating in LocationDetails the logic of which categories start out expanded, and keeping this component focused more on what that looks like.

It makes no functional difference, just a matter of naming semantics: "isSearchCategory" means CategoryCard knows that the the searched card should start out expanded, whereas a name like "isInitiallyExpanded" or "defaultExpanded" would leave the knowledge of why that is to the component passing the prop.

(This will make more difference once we want to determine expanded cards not just by search category, but also search string etc. An abstract prop that just tells CategoryCard whether to start out expanded would mean only LocationDetails has to change, whereas "isSearchCategory" would require a change in both LocationDetails and CategoryCard.)

src/app/find/locationDetails/CategoryCard.jsx
```diff
@@ -1,150 +1,40 @@
...
+import ShelterService from './ShelterService';
 
-const renderSchedule = (schedule) => {
-  if (schedule.length === 7 &&
-    schedule.every(day => day.opens_at === '00:00:00' && day.closes_at === '23:59:00')) {
-    return 'Open 24/7';
+// this card contains a category (e.g. Food) and displays
+// all services that a shelter provides within that category
+class CategoryCard extends Component {
+  state = {
```
Rovack, line 10: Looks like this state and the `handleCollapse` function are no longer used, so they can be removed and this component changed back into a functional component.

src/app/find/locationDetails/CategoryCard.jsx
```diff
@@ -1,150 +1,40 @@
...
-      .sort((service1, service2) => service2.lastUpdate - service1.lastUpdate)
-      .map((service, i) => renderService(service, i === services.length - 1))}
-  </div>
-);
+          <div className="serviceCategoryHeaders">{category}</div>
+        </div>
+        <div>
+          {services
+            .sort((service1, service2) => service2.lastUpdate - service1.lastUpdate)
+            .map(service => <ShelterService service={service} selectedCategory={false} key={service.id} showOnMount={isSearchCategory} />)}
```
Rovack, line 33: `selectedCategory` isn't actually used (which makes sense, as any prop that can only get 1 value isn't likely to be needed).

src/app/find/locationDetails/locationDetails.css
```diff
@@ -18,24 +18,37 @@
...
   background-color: #f8f8fc;
   border: 1px solid #dadada !important;
-  padding: 4vw 4vw 4vw 4vw;
+  border-radius: 2px;
   box-shadow: 0vw 2vw 2vw rgba(0,0,0,0.15) !important;
   margin-bottom: 4vh;
   margin-top: 2vh;
 }
 
+.shelterService:last-of-type {
```
Rovack, line 27: Nice! Much better to use `:last-of-type` and keep styling logic entirely in CSS files, than the previous way which had the React code keep track of the last service and inline its style in JSX.

src/app/find/locationDetails/ShelterService.jsx
```diff
@@ -0,0 +1,153 @@
+import React, { useState } from 'react';
```
Rovack, line 1: One general comment is that shelters are only one of the many service categories we have.
Streetlives also supports food providers, clothing, various types of personal care, legal services, health, mail, etc etc etc.

So, a more correct name would be maybe something like "ServiceSection".

src/app/find/locationDetails/ShelterService.jsx
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    return `${dayRanges.map(formatRange).join(', ')} ${hoursString}`;
+  });
+
+  return `Open ${groupStrings.join(', ')}`;
+};
+
+function ShelterService({ service, selectedCategory, showOnMount }) {
+  const [isShown, setIsShown] = useState(showOnMount);
```
Rovack, line 60: Nitpick, but maybe all these instances of "shown" would be more accurate as "expanded".
I mean, the service is _shown_ either way - just the header instead of the full contents when it's collapsed.

Rovack, line 60: Nice use of hooks. I believe it's the first instance of them in this app. :)

src/app/find/locationDetails/locationDetails.css
```diff
@@ -18,24 +18,37 @@
...
+  border-bottom: none;
+}
+
+.shelterService i {
```
Rovack, line 31: I'd tend to avoid giving custom styles to the actual `i` elements within the `Icon` component. It can break consistency between different instances of the atomic components if their styles are manipulated directly rather than through the standardized props.

Would it be possible to just wrap the `Icon` in a `span` or `div` and style those instead?

src/components/icon/Icon.jsx
```diff
@@ -15,6 +15,7 @@ function Icon({
 }) {
   const classNames = cx(circle ? {} : className, {
     Icon: true,
+    'fa-sm': size === 'sm',
```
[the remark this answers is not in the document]
Vikariusu: The smaller icon better matches the design spec

src/app/find/mapPage/MapPage.jsx
```diff
@@ -214,13 +214,13 @@ export default class MapPage extends Component {
...
         <div>Loading results...</div>
       ) : (
-        <div>
-          Showing results for
-          <span className="font-weight-bold ml-1">
-            {this.getCurrentFilterString()}
-          </span>
-        </div>
-      )}
+          <div>
```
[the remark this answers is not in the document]
Vikariusu: Editor automatically moved indents for some reason, reverting

src/app/find/locationDetails/LocationDetails.jsx
```diff
@@ -175,6 +191,7 @@ class LocationDetails extends Component {
 
   render() {
     const { location, locationError, goBack } = this.props;
+    const searchCategory = this.props.match.params.categoryName;
```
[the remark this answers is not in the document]
Vikariusu: Good tip!

src/app/find/mapPage/MapPage.jsx
```diff
@@ -220,7 +220,7 @@ export default class MapPage extends Component {
             {this.getCurrentFilterString()}
           </span>
         </div>
-      )}
+        )}
```
Rovack, line 223: Odd that this change alone remained, but whatever. :)

Vikariusu: fixed!

src/components/icon/Icon.jsx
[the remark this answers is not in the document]
Rovack: Makes sense. This isn't really a problem, but I do think it's a good opportunity to mention something about how we generally approach components:

Since we try to adhere more or less to "atomic design" principles (or just, generally, consistent design...), whenever we see an existing standard component whose props don't seem to match some designs we got, we first push back a bit on the designs to make sure it's intentional.

Specifically, since there are a lot of icons throughout the app, and so far the existing sizes have been sufficient, it means that if the designs for this one are different, then they must be inconsistent with the designs for all existing icons (we even already have several chevrons in the app being used for this exact same purpose).
At that point, the first thing we would typically do is ask the designer whether they intentionally made this icon special for some reason.

If they did, then adding a prop is absolutely correct.
If they didn't, there are 2 options:
1. The existing sizes are all a bit off from the designs
2. The designs in this case accidentally gave it the wrong size

If it's option 1, and the default-sized icons are all a bit too big, we should fix that, and make the default a bit smaller.
If it's option 2, this this icon can just be made larger.

In this case, not passing a size at all results in a height of 13.6, and sm is 11.2.
The designs have a height of 13.
So, the default size is actually closer to the designs than sm.


**TL;DR**: Again, I don't think this case is actually a problem, but I do think it's important in general to maintain consistency.
In this case, personally I probably would revert the change to Icon.jsx and go with the default size, but if you think it looks better this way, it doesn't matter much to me.

Vikariusu:
> y remaining issue is that, as mentioned, it would've been nice for this to be a PR to master, not feature/coronavirus-sstt, so that we don't lose these changes once the outbreak is over.
> But, again, if you think moving the changes onto master and _then_ merging into feature/coronavirus-sstt w

Sounds great. I'll merge for now and make a cherry-picked PR to master as well (there would be some conflicts now as there is a little coronavirus specific code in this PR

PR #150, opened 2020-05-15
Improve proofs required, to support multiple options

abimaelmartell:
Support multiple proofs, and show initial options.

### DIFF
--- a/src/app/team/documents/proofs/DocsProofs.jsx
+++ b/src/app/team/documents/proofs/DocsProofs.jsx
@@ -25,7 +25,7 @@ const mapStateToProps = (state, ownProps) => ({
 const mapDispatchToProps = (dispatch, ownProps) => ({
   fetchResourceData: bindActionCreators(actions.getLocation, dispatch),
   updateValue: (proofs, serviceId, metaDataSection, fieldName) => {
-    if (proofs.every(p => (typeof p !== 'string'))) {
+    if (proofs.length === 0) {
       return null;
     }
 
--- a/src/app/team/documents/proofs/DocsProofs.jsx
+++ b/src/app/team/documents/proofs/DocsProofs.jsx
@@ -2,45 +2,45 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { compose, withProps } from 'recompose';
-
-import { getDocuments, getFirstProof } from '../../../../selectors/document';
+import { getDocuments, getRequiredDocuments } from '../../../../selectors/document';
 import { getServiceId } from '../../../../selectors/service';
 import { selectLocationError } from '../../../../selectors/location';
-
 import * as actions from '../../../../actions';
-import { Form, FormEdit, FormView } from '../../../../components/form';
-
-const EditComponent = compose(withProps({
-  headerText: 'What proofs are required to use this service?',
-  placeholderText: 'e.g. Proof of identity, proof of address',
-}))(props => <FormEdit {...props} />);
-
-const ViewComponent = compose(withProps({
-  topText: 'PROOFS REQUIRED',
-}))(props => <FormView {...props} />);
+import { Form } from '../../../../components/form';
+import Edit from './Edit';
+import View from './View';
 
 const FormComponent = compose(withProps({
-  ViewComponent,
-  EditComponent,
+  ViewComponent: View,
+  EditComponent: Edit,
 }))(props => <Form {...props} />);
 
 const mapStateToProps = (state, ownProps) => ({
   resourceData: getDocuments(state, ownProps),
-  value: getFirstProof(state, ownProps),
+  value: getRequiredDocuments(state, ownProps),
   id: getServiceId(ownProps),
   resourceLoadError: selectLocationError(state, ownProps),
 });
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
   fetchResourceData: bindActionCreators(actions.getLocation, dispatch),
-  updateValue: (value, serviceId, metaDataSection, fieldName) =>
-    dispatch(actions.updateService({
+  updateValue: (proofs, serviceId, metaDataSection, fieldName) => {
+    if (proofs.every(p => (typeof p !== 'string'))) {
+      return null;
+    }
+
+    return dispatch(actions.updateService({
       locationId: ownProps.match.params.locationId,
       serviceId,
-      params: { documents: { proofs: [value] } },
+      params: {
+        documents: {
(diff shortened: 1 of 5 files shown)


src/reducers/locations.js
```diff
@@ -69,18 +69,15 @@ const locationsReducer = (state = {}, action) => {
...
 
-        if (documents.proofs != null) {
-          // TODO: Update when form support multiple proofs
-          // eslint-disable-next-line prefer-destructuring
-          const proof = { ...RequiredDocuments[0], document: documents.proofs[0] };
+        let { RequiredDocuments } = service;
 
-          RequiredDocuments[0] = proof;
+        if (documents.proofs != null) {
+          RequiredDocuments = documents.proofs.map(p => ({ document: p }));
```
Rovack, line 80: Just to make sure: There are no fields on the object other than `document`, that need to be preserved?
(I see the previous code was keeping all the other fields' values, but don't know if there was a reason for that.)

src/app/team/documents/proofs/DocsProofs.jsx
```diff
@@ -2,45 +2,45 @@ import React from 'react';
...
   id: getServiceId(ownProps),
   resourceLoadError: selectLocationError(state, ownProps),
 });
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
   fetchResourceData: bindActionCreators(actions.getLocation, dispatch),
-  updateValue: (value, serviceId, metaDataSection, fieldName) =>
-    dispatch(actions.updateService({
+  updateValue: (proofs, serviceId, metaDataSection, fieldName) => {
+    if (proofs.every(p => (typeof p !== 'string'))) {
```
Rovack, line 28: So looks like this is checking if the proofs array has at least 1 real value in it.
That makes sense, but then what will happen with any values that are _not_ strings? If they're still sent to the action without any filtering, will they end up deleting existing documents, or...?

src/app/team/documents/proofs/Edit.jsx
```diff
@@ -0,0 +1,160 @@
...
+  'Proof of address',
+  'Proof of income',
+  'Birth certificates',
+];
+
+class ProofsRequiredEdit extends Component {
+  constructor(props) {
+    super(props);
+
+    let proofs;
```
Rovack, line 20: Doesn't this need to be initialized to `[]`?
If `this.props.value` is falsey, `proofs` will be undefined, and then `...proofs` will throw a TypeError.

src/app/team/documents/proofs/Edit.jsx
```diff
@@ -0,0 +1,160 @@
...
+  render() {
+    const { state: { proofs, showForm } } = this;
+
+    const totalProofs = [...new Set([
+      ...DEFAULT_PROOFS,
+      ...proofs,
+    ])];
+
+    return (
+      <div className="w-100 WhoDoesItServe">
```
Rovack, line 121: Does this actually need styles from `WhoDoesItServe`?
If so, those styles should probably be renamed, and the relevant CSS should be imported.

src/app/team/documents/proofs/Edit.jsx
```diff
@@ -0,0 +1,160 @@
...
+    super(props);
+
+    let proofs;
+
+    if (this.props.value) {
+      proofs = this.props.value.map(el => el.document);
+    }
+
+    this.state = {
+      proofs,
```
Rovack, line 27: It's [kind of an anti-pattern](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html) to copy state into props, and especially so only in the constructor (since prop changes are ignored).
However, I don't think it's that big of a problem here, so this is fine.

src/selectors/document.js
```diff
@@ -5,9 +5,8 @@ export const getDocuments = (state, props) => getService(state, props).Documents
 export const getRequiredDocuments = (state, props) =>
   getService(state, props).RequiredDocuments || [];
 
-// TODO Replace whenever form supports multiple proofs
-export const getFirstProof = (state, props) =>
-  (getRequiredDocuments(state, props)[0] || {}).document;
+export const getRequiredProofs = (state, props) =>
+  (getService(state, props).documents || {}).proofs || [];
```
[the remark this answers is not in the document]
Rovack: Though now that I look at it again, I see we actually do have this notion of "proofs", at least on the updating side. I still think that's weird, but if it's existing weirdness, that's not so bad. :)

src/reducers/locations.js
abimaelmartell: Yeah, it's only taking that string on the backend.

src/selectors/document.js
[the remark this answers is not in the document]
abimaelmartell: Yeah, i added it because i saw that on the backend

src/app/team/documents/proofs/Edit.jsx
abimaelmartell: Copy paste error 😅

src/app/team/documents/proofs/DocsProofs.jsx
abimaelmartell: I think it was meant to be a null check, but is not required

src/app/team/documents/proofs/Edit.jsx
abimaelmartell: Yeah, this doesn't use any kind of props change, that's why i added it to the constructor only

abimaelmartell:
@Rovack updated

src/app/team/documents/proofs/DocsProofs.jsx
Rovack: Oh, OK. I hope that was indeed the only edge case this was meant to cover.

src/app/team/documents/proofs/Edit.jsx
```diff
@@ -0,0 +1,160 @@
...
+      ],
+      showForm: false,
+    }));
+  }
+
+  onProofClick(name) {
+    let { proofs } = this.state;
+
+    if (name === 'None') {
+      proofs = ['None'];
```
Rovack, line 56: I must have missed it in the review, but do we actually end up sending "None" to the backend, and storing it in the DB, or is it filtered out at some point?
If so, I think that's a good idea (so we can distinguish unknown proofs from known-to-be-empty proofs), but given our current user-facing code doesn't know about this, we might end up showing the user something like "Requires None".

So, if that is the case, please just let me know, so I can make a ticket for updating the user-facing page accordingly.

PR #151, opened 2020-05-16
Feature/handle no results

Vikariusu:
Asana ticket - https://app.asana.com/0/1168502694533225/1174572971197995
Includes a demo of the feature and the map drag fix.

### DIFF
--- a/src/app/find/mapPage/MapPage.jsx
+++ b/src/app/find/mapPage/MapPage.jsx
@@ -24,7 +24,7 @@ const initialFiltersState = {
 export default class MapPage extends Component {
   state = {
     initialLocationsLoaded: false,
-    locations: [],
+    locations: null,
     zoomedLocations: null,
     center: null,
     radius: null,
@@ -32,6 +32,7 @@ export default class MapPage extends Component {
     isSearchingLocations: false,
     isFilterModalOpen: false,
     filters: initialFiltersState,
+    hasResults: false,
   };
 
   componentDidMount() {
@@ -193,6 +194,7 @@ export default class MapPage extends Component {
     this.setState({ isSearchingLocations: true }, () => {
       this.fetchLocations(minSearchResults)
         .then(() => this.setState({
+          hasResults: this.state.locations && this.state.locations.length,
           isSearchingLocations: false,
           zoomedLocations: this.state.locations && this.state.locations.slice(0, minSearchResults),
         }));
@@ -306,9 +308,9 @@ export default class MapPage extends Component {
                   <ResultsBar
                     isSearching={this.state.isSearchingLocations}
                     filterString={this.getCurrentFilterString()}
-                    hasResults={!!this.state.locations.length}
+                    hasResults={this.state.hasResults}
                     filters={this.state.filters}
-                    clearResults={this.clearResults} 
+                    clearResults={this.clearResults}
                   />
                 ) : renderSearchBar()
               }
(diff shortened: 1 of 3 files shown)


src/app/find/mapPage/ResultsBar.jsx
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  noResultsInfo = () => {
+    const { hasResults, filters, clearResults } = this.props;
+    const searchTerm = filters.searchString;
+
+    // No results after searching through the searchbar
+    if (searchTerm && searchTerm.length && !hasResults) {
+      const wordCount = searchTerm.split(' ').length;
```
Rovack, line 23:
```
> '       hi     there    '.split(' ').length
17
```

I'd go with something more like `searchTerm.trim().split(/\s+/)`.

src/app/find/mapPage/ResultsBar.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+    if (isSearching) {
+      // loading state
+      return <div>Loading results...</div>;
+    } else if (hasResults) {
+      // results found
+      return (
+        <div>
+          Showing results for
+          <span className='font-weight-bold ml-1'>{filterString}</span>
```
Rovack, line 82: JSX attribute values use double quotes.

I don't actually care about that, but my concern is that eslint shows an error for this, which means your eslint likely isn't working.
In streetlives-api eslint checking is enforced through a pre-commit hook, but for streetlives-web it's important to enable it in your IDE, as otherwise you might not see the warnings.

src/app/find/mapPage/ResultsBar.jsx
```diff
@@ -0,0 +1,96 @@
...
+        </div>
+      );
+    }
+  };
+
+  displayContent = () => {
+    const { isSearching, hasResults, filterString } = this.props;
+
+    if (isSearching) {
+      // loading state
```
Rovack, line 75: IMO when the line is `return <div>Loading results...</div>;`, then "loading state" is pretty self-evident 😛.
(Even truer for `hasResults` -> `// results found`.)

This is actually great - it means your code and naming choices are readable enough to be self-documenting, which is what all code should strive for.
This is why nowadays a lot of teams actively discourage comments in the code, as well-written code (like this) shouldn't need them in the vast majority of cases.

src/app/find/mapPage/MapPage.jsx
```diff
@@ -23,7 +24,7 @@ const initialFiltersState = {
 export default class MapPage extends Component {
   state = {
     initialLocationsLoaded: false,
-    locations: null,
+    locations: [],
```
Rovack, line 27: Starting with an empty array means we don't necessarily have a way to distinguish the case where no search has happened yet from the case a search has happened, but returned nothing. That could e.g. lead to a brief flash of the "no results" bar when the page loads, if it renders before the first search is dispatched.

I don't think it matters much here, but generally it can help to distinguish _not having data_ from having data and _knowing it's empty_, and null can be useful for the former.

src/app/find/mapPage/MapPage.jsx
```diff
@@ -306,9 +308,9 @@ export default class MapPage extends Component {
                   <ResultsBar
                     isSearching={this.state.isSearchingLocations}
                     filterString={this.getCurrentFilterString()}
-                    hasResults={!!this.state.locations.length}
+                    hasResults={this.state.hasResults}
```
Rovack, line 311: This is actually an integer rather than a boolean, since it's `hasResults: this.state.locations && this.state.locations.length` without `!!`, but that doesn't actually matter.

Rovack:
Oh, and I see it's into master - perfect!
Just remember to merge master into feature/coronavirus-sstt right after you merge this.

PR #153, opened 2020-05-17
Feature/blue icon if active gray if not

AdamNgu:
This actually also contains my changes where the FE requests limited location data on the /find part of the site, I forgot to push and PR that earlier.

### DIFF
--- a/src/app/find/mapPage/MapPage.jsx
+++ b/src/app/find/mapPage/MapPage.jsx
@@ -162,6 +162,7 @@ export default class MapPage extends Component {
       radius: Math.floor(radius),
       minResults,
       searchString,
+      basicMapOnly: true,
       serviceFilters: {
         ...eligibilityParams,
         ...filtersObject,
--- a/src/app/find/mapPage/MapPage.jsx
+++ b/src/app/find/mapPage/MapPage.jsx
@@ -162,7 +162,7 @@ export default class MapPage extends Component {
       radius: Math.floor(radius),
       minResults,
       searchString,
-      basicMapOnly: true,
+      locationFieldsOnly: true,
       serviceFilters: {
         ...eligibilityParams,
         ...filtersObject,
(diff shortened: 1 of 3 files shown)


src/app/find/mapPage/MapPage.jsx
```diff
@@ -162,6 +162,7 @@ export default class MapPage extends Component {
       radius: Math.floor(radius),
       minResults,
       searchString,
+      locationFieldsOnly: true,
```
Rovack, line 165: Wasn't this change already made in the last PR?

src/components/map/LocationMarker.jsx
```diff
@@ -2,6 +2,11 @@ import React from 'react';
 import { Marker, InfoWindow } from 'react-google-maps';
 import Button from '../button';
 
+const MARKER_ICON_MAPPINGS = {
+  blue: 'https://www.google.com/maps/vt/icon/name=assets/icons/spotlight/spotlight_pin_v2_shadow-1-small.png,assets/icons/spotlight/spotlight_pin_v2-1-small.png,assets/icons/spotlight/spotlight_pin_v2_dot-1-small.png,assets/icons/spotlight/spotlight_pin_v2_accent-1-small.png&highlight=000000,1400FF,FFFFFF,000000&color=1400FF?scale=1',
```
Rovack, line 6: Just to make sure, this is the official way Google exposes these markers? No CDN or anything?

src/app/find/mapPage/MapPage.jsx
AdamNgu: I think I forgot to make that PR for the streetlives-web repo :x so here it is

src/components/map/LocationMarker.jsx
AdamNgu: Correct, this is the official way to have custom colors on markers

PR #154, opened 2020-05-17
Feat/landing page mvp

Vikariusu:


src/app/landing/LandingPage.jsx
```diff
@@ -0,0 +1,258 @@
...
+          <GetStartedButton onClick={this.getStarted} fluid className="btnWhite" />
+        </div>
+        <div className="Partners Section">
+          <h3>Collaboration</h3>
+          <h2>Love for the help we’ve had to create GoGetta!</h2>
+          <div className="PartnerPictures">
+            <PartnerPicture
+              link="http://www.aliforneycenter.org/"
+              alt="Ali Forney logo"
+              src="img/partners/ali_forney_logo.png"
```
Rovack, line 140: If these were moved to the /partners dir, doesn't that mean the /about page is now broken?
Not that I expect anyone to use it right now, but I wouldn't want a completely broken page. Maybe just fix the links there too, if it's easy?

AdamNgu: Fixed

PR #157, opened 2020-05-18
Feature/error report

cadecameron:
Hey @Rovack ,

Take a look at the following pull request, which sets-up the frontend error reporting workflow under the `find` route.

Note: I haven't added a button to navigate to the beginning of the error reporting flow yet. I'll do that small step in a separate PR.


src/app/find/errorReports/ErrorReportContainer.jsx
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  onErrorReportTextChanged(event) {
+    this.setState({ errorReportText: event.target.value });
+  }
+
+  onErrorReportServicesChanged(value) {
+    const newErrorReportInformation = { ...this.state.errorReportInformation };
+
+    const newErrorReportServices = newErrorReportInformation.errorReportServices.includes(value)
```
Rovack, line 55: This is the kind of low-level logic I'd keep to the specific pages' components. Otherwise, this container doesn't just concern itself with the high-level flow, but basically has all the logic for every level of the flow in it.
Ideally, its function would just be:
```
onErrorReportServicesChanged(errorReportInformation) {
  this.setState({ errorReportInformation });
}
```
And the child components would make sure to do everything else.

But if it's not a really quick fix, I wouldn't bother with it right now.

src/app/find/errorReports/ErrorReportContainer.jsx
```diff
@@ -0,0 +1,187 @@
...
+                onClick={goToViewLocation}
+                style={{
+                  position: 'absolute',
+                  right: 0,
+                  top: '0.2em',
+                }}
+              />
+            </div>
+            <Switch>
+              <Redirect exact from={`${this.props.match.path}/`} to={`${this.props.match.path}/services`} />
```
Rovack, line 125: Is the redirection actually needed? Why not just make services be at `/` in the first place?
(But this isn't really an issue, unless it affects back navigation in some weird way, which it probably doesn't.)

cadecameron: I did consider services just being at `/`, but thought I'd better give it it's own named endpoint.
Can change if you'd like, or leave for a refactor later.

src/app/find/errorReports/ErrorReportInformationSelect.jsx
```diff
@@ -0,0 +1,98 @@
...
+                checked={errorReportServices.includes(service.id)}
+                onChange={() => onServiceChange(service.id)}
+              />
+            </div>
+          ))}
+        </div>
+      </div>
+
+      <div className="mx-4 mb-1">
+        <Link to={`${props.match.url}/text`}>
```
Rovack, line 52: Hmm, so despite the huge amount of logic concentrated in in the container, the URL schema, i.e. the high-level flow of the pages, is actually the one thing that _isn't_?

I'd try to at least keep that encapsulated there.
Just means passing a function prop to this component, and then invoking it on the button, instead of wrapping it in a Link (!?) and making it do nothing.

src/app/find/errorReports/ErrorReportText.jsx
```diff
@@ -0,0 +1,78 @@
...
+            value={errorReportText}
+            onChange={onChange}
+            minRows={7}
+            autoFocus
+            fluid
+            className="border"
+          />
+        </div>
+        <div className="mx-4 mb-1">
+          <Link to={`${props.match.url}/thanks`}>
```
Rovack, line 40: Same comment as in the other component: This is really the 1 thing that _should_ be handled by the container, rather than having a Button that doesn't really do anything wrapped in a link.

It just never would have occurred to me to look here for how navigation happens, when in every other respect this is entirely a presentational component.

src/app/find/errorReports/ErrorReportInformationSelect.jsx
cadecameron: Yeah, that's the original way I had it, however I was running into issues where the URL would update, but not render the new Route.

src/app/find/errorReports/ErrorReportContainer.jsx
```diff
@@ -0,0 +1,187 @@
...
+    </LoadingLabel>
+  </div>
+);
+
+class ErrorReportContainer extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      errorReportInformation: {
+        errorReportGeneralLocationError: false,
```
Rovack, line 30: Since you're initializing this with valid errorReportGeneralLocationError and errorReportServices values, didn't you say that means you can't distinguish the case where no data has been entered from the one where it was entered, and was just empty?

Well, anyway, it's fine for now. Just means that if a user refreshes while in this flow, they'll submit without knowing their checkbox selections are wiped, and we'll get the data as if they left them empty. But that should be a rare case, hopefully.

src/app/find/errorReports/ErrorReportInformationSelect.jsx
cadecameron: Originally all navigation was like the `goToViewLocation` function in `mapDispatchToProps` in the `ErrorReportContainer.jsx` file.

src/app/find/errorReports/ErrorReportContainer.jsx
cadecameron: Agree.
Yeah, the use-case where a user is currently navigating straight to `/thanks` or `/text` isn't being handled currently. I've got the object in place, and it should be initialized with null going forward so these checks can occur (I didn't get around to implementing the logic).

src/app/find/errorReports/ErrorReportInformationSelect.jsx
cadecameron: I'll take another look at what I might have been doing wrong previously.

cadecameron:
This should be good to go after a final review of the last commmit @Rovack

src/app/find/errorReports/ErrorReportInformationSelect.jsx
```diff
@@ -49,19 +49,17 @@ const ErrorReportInformationSelect = (props) => {
...
-              (!generalLocationError && !errorReportServices.length)
-                ? 'Skip'
-                : 'Next'
-            }
-          </Button>
-        </Link>
+        <Button
+          primary
+          fluid
+          onClick={goToErrorReportTextScreen}
```
Rovack, line 55: I'd rename this to `onSubmit` too, as it''s named in the other page.
Since we're making these components purely presentational, they don't need to know what happens when they're finished - that's up to the container, and they just need to call the function it gave them.

src/app/find/errorReports/ErrorReportContainer.jsx
```diff
@@ -107,65 +110,67 @@ class ErrorReportContainer extends Component {
...
+              exact
+              from={`${this.props.match.path}/`}
+              to={`${this.props.match.path}/services`}
+            />
+            <Route
+              exact
+              path={`${this.props.match.path}/text`}
+              render={() => (
+                <ErrorReportText
+                  match={this.props.match}
```
Rovack, line 137: No need to pass `match` anymore, right?
After all, the child components don't deal with navigation and URL concerns anymore.

cadecameron:
Made those requested changes @Rovack

PR #161, opened 2020-05-18
feat: Add coronavirus modal on landing page

Vikariusu:


src/app/landing/CovidInfoModal.jsx
```diff
@@ -0,0 +1,46 @@
...
+    e.preventDefault();
+    history.push('/find');
+  };
+
+  return (
+    <Modal compact>
+      <div className="covidTopBar">
+        <div className="m-4"><img src="/img/virus.png" alt="virus icon" /></div>
+        <div onClick={closeModal} className="closeBtn">
+          <Icon name="times" />
```
Rovack, line 19: The up-to-date designs don't include a way to close the modal, and I believe that's because once the user has clicked "Get Started", the thinking is that they're already "in transition" to the map.
In fact, it would've probably made sense for the modal to be on top of the map and not the landing page (so the locations can load in the background while the user is reading it), but whatever.

Anyway, I don't think there's a reason to let the user go back, as if they've already clicked the CTA, we want them to get into the actual app. And of course, once they're there, they can always just navigate back.

If you feel differently, though, that could be discussed with Di.

src/app/landing/CovidInfoModal.jsx
```diff
@@ -0,0 +1,46 @@
...
+        <div>
+          <p>During the pandemic we will show when information on every location was last updated, but please contact providers before accessing their services.</p>
+          <p>At this time some of our regular features may not be available.</p>
+        </div>
+        <div>
+          <div className="px-3">
+            <Button primary fluid onClick={toMap}> Great, continue </Button>
+          </div>
+          <div className="text-center legal">
+            <a href="https://docs.google.com/document/d/e/2PACX-1vTBNI2Sv5QC8DSBwBL7WHNBdMI-9kPLuN2ev_Y2VDSo-bLeh8qssi7iBv-w0EEQurX9fgFQF4_2lItn/pub" target="_blank" rel="noopener noreferrer">Terms of Use</a>
```
Rovack, line 36: I was going to say that usually I'd store values like this in config.js.
But then I opened config.js, and saw that they're already there. :P

I haven't checked whether the links are exactly identical, but if they're not, then just replace the old values with these.

PR #167, opened 2020-05-21
fix: optimize for mobile

Vikariusu:
Swapped desktop and mobile images for optimized images provided by the designer to improve network performance.


src/app/landing/Testimonial.css
```diff
@@ -11,9 +11,9 @@
 }
 
 .Testimonial-image {
-  border-radius: 100%;
+  /* border-radius: 100%;
```
Rovack, line 14: I assume this is no longer needed because the images themselves were edited to match this style?
It's generally preferable to leave stylistic cropping up to CSS, as opposed to "hard-coding" the image a certain way. That way, any stylistic changes don't require re-generating all the pictures (assuming the originals are even still around).

Rovack, line 14: I assume this is no longer needed because the images themselves were edited to match this style?
It's generally preferable to leave stylistic cropping up to CSS, as opposed to "hard-coding" the image a certain way. That way, any stylistic changes don't require re-generating all the pictures (assuming the originals are even still around).

Or did the designs actually change, s.t. we no longer want these to be round?

src/app/landing/PartnerPicture.jsx
```diff
@@ -1,12 +1,33 @@
...
 
-const PartnerPicture = ({ src, alt, link }) => (
+const PartnerPicture = ({ img, alt = '', link }) => (
   <div className="PartnerPicture">
     <a href={link}>
-      <img src={src} alt={alt} />
+      <picture>
+        <source
+          className="OnboardingImage"
+          media="(min-aspect-ratio: 2/3)"
```
Rovack, line 11: For the map picture, we used the ratio as the deciding factor because it was the reason we had 2 versions in the first place: centering the markers on the map in both "portrait" and "landscape" displays, while keeping it full-width, was difficult with a single version.

However, since the factor that makes us use 2 versions in this case is actually the resolution/display size, the breakpoint should be based on that as well, and not on the ratio (while it's a good hueristic that phones are more "vertical" than desktops, a phone rotated sideways, for example, still doesn't need a large version of a picture).

(That of course goes for all the new assets, not just this place.)

src/app/landing/LandingPage.jsx
```diff
@@ -136,87 +160,138 @@ class LandingPage extends Component {
...
+              img={{
+                small: 'img/partners/openreferral_mobile_logo.png',
+                large: 'img/partners/openreferral_logo.png',
+              }}
             />
             <PartnerPicture
               link="https://holyapostlessoupkitchen.org/"
               alt="Holy Apostles logo"
-              src="img/partners/holy_apostles_logo.png"
+              img={{
```
Rovack, line 275: An alternative solution, that wouldn't require so many changes to the code (and won't require further changes if we need to change this again), would be to do something like what native apps do: put different sizes of the same picture in a directory, or some standardized naming format, pass to the component just the base name (e.g. holy_apostles_logo.png) and then let the component handle the logic of choosing a version.

But not necessary, if you think it would take non-trivial work.

src/app/landing/LandingPage.jsx
```diff
@@ -99,10 +109,15 @@ class LandingPage extends Component {
...
+                className="OnboardingImage"
+                media="(min-aspect-ratio: 2/3)"
+                srcSet="img/landing_page/step3.png"
+              />
               <img
                 className="OnboardingImage"
-                src="img/landing_page/step3.png"
-                alt=""
+                src="img/landing_page/step3_mobile.png"
+                alt="map showing food, clothing, and personal care markers"
```
Rovack, line 120: It's also supposed to show shelter IIRC, unless we never updated it.

src/app/landing/LandingPage.css
```diff
@@ -248,4 +248,12 @@
   .LandingPage h3 {
     font-size: 11px;
   }
+
+  .LandingPage .IntroBody {
+    background: linear-gradient(to bottom, rgba(0, 0, 0, 0), rgba(0, 0, 0, 0.6)),url('/img/landing_page/top1_mobile.png') no-repeat center center fixed;
+    background-position: 30% 0%;
+    background-size: cover;
+  /* color: #FFFFFF; */
```
Rovack, line 256: Why the duplication?

src/app/landing/LandingPage.jsx
```diff
@@ -112,17 +127,26 @@ class LandingPage extends Component {
           <div className="TestimonialsInner">
             <h3 className="TestimonialsTitle">GoGettas are saying</h3>
             <Testimonial
-              img="img/landing_page/testimonial_jeffrey.png"
+              img={{
+                small: 'img/landing_page/testimonial_jeffrey_mobile.png',
```
Rovack, line 131: If GitHub is to be trusted, these testimonials on mobile aren't actually smaller than the normal versions (~25KB in both cases).
In general, worth reviewing all the file sizes and seeing the mobile is actually an improvement in all cases.

src/app/landing/LandingPage.jsx
```diff
@@ -136,87 +160,138 @@ class LandingPage extends Component {
             <PartnerPicture
               link="http://www.aliforneycenter.org/"
               alt="Ali Forney logo"
-              src="img/partners/ali_forney_logo.png"
+              img={{
```
Rovack, line 163: I recall AdamB and Di were talking about how on mobile, these pictures should just be the icons, with no text (on a small screen, it's basically impossible to read the text anyway). Perhaps you should ask her if we have those assets, so that we could make that change now already.

src/app/landing/LandingPage.jsx
Rovack, line 120: Wait, more importantly... step **1** is the map, isn't it? This is just a picture of a person.
(I think it might still be a good idea to still describe what it shows, though `alt=""` is acceptable for purely decorative pictures. Whatever you think.)

src/app/landing/LandingPage.jsx
```diff
@@ -65,26 +65,36 @@ class LandingPage extends Component {
               <source
                 className="OnboardingImage"
                 media="(min-aspect-ratio: 2/3)"
-                srcSet="img/landing_page/map_tablet.png"
+                srcSet="img/landing_page/step1.png"
```
Rovack, line 68: Looking at the video, I see the Food and Shelter markers are truncated. Has that always been the case?

src/app/landing/LandingPage.jsx
```diff
@@ -65,26 +65,36 @@ class LandingPage extends Component {
...
               />
               <img
                 className="OnboardingImage"
-                src="img/landing_page/map_narrow.png"
+                src="img/landing_page/step1_mobile.png"
```
Rovack, line 72: In the video, this map (and maybe a couple other places) seem a bit blurry. Is that just in the video, or does the resolution/quality for the mobile versions need to be improved?
If the latter, maybe Di could help.

src/app/landing/LandingPage.jsx
```diff
@@ -65,26 +65,36 @@ class LandingPage extends Component {
...
               <p className="text-lighter">
                 Filter or answer a few questions and we will show you locations that meet your specific needs, from age to gender to family situations.
               </p>
             </div>
-            <img className="OnboardingImage" src="img/landing_page/welcome.jpg" alt="" />
+            <picture>
+              <source
+                className="OnboardingImage"
+                media="(min-aspect-ratio: 2/3)"
+                srcSet="img/landing_page/step2.png"
```
Rovack, line 90: One of the other notes highlighted red in the web.dev audit was that we should be using "next-gen formats" for these pictures, rather than pngs and jpgs. Not necessary to do it as part of this PR, but since we're already replacing these assets, maybe it would be simpler?

src/app/landing/LandingPage.jsx
```diff
@@ -99,10 +109,15 @@ class LandingPage extends Component {
             </div>
 
             <picture>
+              <source
+                className="OnboardingImage"
+                media="(min-aspect-ratio: 2/3)"
+                srcSet="img/landing_page/step3.png"
```
Rovack, line 115: Woah, how did the file size for this step go up by 950%?
Unless GitHub is just confused, it claims this and step2, on the desktop versions, are now over 1MB in size. Is that intentional?

src/app/landing/LandingPage.jsx
Vikariusu: It's the latter, these were the images provided for mobile resolutions. I sent her a message, maybe we can find a middle ground.

src/app/landing/LandingPage.jsx
Vikariusu: Following up with Di to reduce image sizes

src/app/landing/Testimonial.css
Vikariusu: That's correct, the testimonial images provided to me are already round, so this CSS was causing the halo effect to double. I'll reach out for compressed but unedited images.

src/app/landing/PartnerPicture.jsx
Vikariusu: Sounds good, do you or Di have a preference for a specific breakpoint to use for mobile?

If not, I will go ahead and try

/* Smartphones (portrait) ----------- */
@media only screen and (max-width : 320px) {
/* Styles */
}
as set by https://responsivedesign.is/develop/browser-feature-support/media-queries-for-common-device-breakpoints/

src/app/landing/LandingPage.jsx
Vikariusu: Good idea, I'll change the filenames and implement this.

src/app/landing/LandingPage.jsx
Vikariusu: Will add!

src/app/landing/LandingPage.css
Vikariusu: I'm experiencing a CSS bug.
Without duplicating the background-position and background-size, I get this:
image

I think setting background on this more specific selector overrides/ resets background-position and background-size.
Also removed the commented out CSS

src/app/landing/LandingPage.jsx
Vikariusu: oops, fixed

src/app/landing/LandingPage.jsx
Vikariusu: The alternative is to look like this:
image

I will ask Di for what she prefers

src/app/landing/PartnerPicture.jsx
Rovack: I'm not actually sure it should be based on height/width, as opposed to resolution ("device-pixel-ratio" or whatnot). Looks like the `picture` element already has `srcset` with the ability to specify different pictures for "1x", "2x" etc, which sounds more correct.

But, I haven't looked into it too much. There's probably posts out there with best practices, so I'd just find a reliable one and follow whatever it says.

src/app/landing/LandingPage.css
Rovack: Hmm, but the selector looks exactly the same: it's `.LandingPage .IntroBody` at the top of the file too.
It's probably something small we're missing... But the duplication, with the exact same selector, can't possibly be necessary.

src/app/landing/LandingPage.jsx
Rovack: You mean there'd be white padding at the bottom? That's almost certainly worse than some truncation.
However, I believe the previous versions of the pictures somehow managed to avoid both, thanks to having the right ratios and "padding" in the 2 different versions...
Maybe the ratios are different this time?

Probably best to try prod/test for a second with whatever device size you're using, and see if it has the same problem. If it does, then maybe we just need to accept this.

Vikariusu:
I addressed all the issues except the next-gen image formats and some other assets. 
I made a ticket grouping up all assets-related desired updates here: https://app.asana.com/0/1168502694533225/1178203231455937 (most notably a browser lack of support for the jpf format)

As of now, this PR doesn't address the biggest image issue (2 files above 1 MB) so I can either wait to get those new assets or close this PR for now and add them on a new PR when we receive the updated images.

Rovack:
I think it's OK to merge this without all the asset size reductions. Thanks for opening the follow-up tasks.
But can we at least revert the pictures that actually became _larger_ as part of this PR? I mean, if we already had an image that was just ~200KB, and now it's somehow 1.5MB, we should just go back to using the previous one, no?

src/app/landing/LandingPage.css
```diff
@@ -41,11 +41,12 @@
   flex-direction: column;
   justify-content: space-between;
   align-items: center;
-  background: linear-gradient(to bottom, rgba(0, 0, 0, 0), rgba(0, 0, 0, 0.6)),url('/img/landing_page/top1.jpg') no-repeat center center fixed;
+  background: linear-gradient(to bottom, rgba(0, 0, 0, 0), rgba(0, 0, 0, 0.6)),url('/img/landing_page/top1_desktop.jpg') no-repeat center center fixed;
   background-position: 30% 0%;
   background-size: cover;
   color: #FFFFFF;
   padding: 15vh 0px 15vh 0px;
+  overflow: hidden;
```
Rovack, line 49: Not a problem, but just out of curiosity, how come this was needed? Aren't all the heights in vh units, meaning they shouldn't be able to overflow on any screen size?

src/app/landing/LandingPage.css
```diff
@@ -233,7 +238,7 @@
 }
 
 @media screen and (max-width: 470px) {
-  .LandingPage {
+.LandingPage {
```
Rovack, line 241: Indentation broke here.

src/app/landing/LandingPage.css
```diff
@@ -249,3 +254,26 @@
     font-size: 11px;
   }
 }
+
+@media screen and (max-width: 1224px) {
+  .LandingPage .IntroBody {
+    background: linear-gradient(to bottom, rgba(0, 0, 0, 0), rgba(0, 0, 0, 0.6)),url('/img/landing_page/top1_mobile.jpg') no-repeat center center fixed;
+    background-position: 30% 0%;
```
Rovack, line 261: Doesn't hurt to have background-position and background-size here, but do they actually make any difference? As far as I can tell they would have these values anyway, thanks to the existing `.LandingPage .IntroBody` styles above, outside the breakpoint.

src/app/landing/LandingPage.css
```diff
@@ -249,3 +254,26 @@
...
+    background-size: cover;
+  }
+}
+
+@media screen and (min-width: 1224px) {
+  .IntroTop, .IntroBottom, .OnboardingText {
+    max-width: 800px;
+  }
+
+  .OnboardingText {
```
Rovack, line 271: Not really sure why OnboardingText needs its own section, considering just putting it in the one below would already give it the margins and max width, but alright.

src/app/landing/LandingPage.jsx
```diff
@@ -57,34 +64,90 @@ class LandingPage extends Component {
...
+              <source
+                media="(min-width: 1224px)"
+                srcSet="/img/landing_page/step1_desktop1x.webp 1x,
+              /img/landing_page/step1_desktop2x.webp 2x"
+                type="image/webp"
+              />
+              {/* mobile jpg */}
+              <source
+                media="(max-width: 1223px)"
+                srcSet="/img/landing_page/step1_mobile2x.jpg"
```
Rovack, line 90: I'm still a bit confused about the sizes: If I understand correctly, currently all devices with a size under 1224 _CSS_ pixels, i.e. _all_ phones, will get this version, which I assume is the one Di provided that is e.g. 1224x817 for this step1 picture.
That seems a little small even for old phones with just 2x density (even e.g. an iPhone 6 has a size of 750x1334 hardware pixels, which, if the picture takes up say 60% of the viewport height - 800px - is just barely small enough not to require stretching).
Certainly if we're talking about newer devices, with better density or larger size than an iPhone 6, the picture won't be as sharp as it could be.

Of course, this isn't a huge deal (the primary users we care about presumably don't have the most cutting-edge devices), but the question is why we wouldn't look at the density for the mobile case too, as you already do for desktop?
In a way, it seems referring to the images as "mobile2x", "desktop1x", and "desktop2x" is somewhat misleading, because in reality a high-density phone could benefit from a larger file size just as much as a normal-density desktop.

I suppose one reason could be that even for high-res phones, there's more of a likelihood that they're using a mobile data connection, meaning bandwidth is a concern, whereas for desktops that's less likely to be the case (although anyone with a high-res phone is more likely to have Unlimited Data anyway...).

But on that note, my understanding is that there's already a way to let the browser make the decision of which picture to choose, based on all available information, including bandwidth: Just use `srcset` with `w` values specifying the file sizes (and pass `sizes`), instead of relying on `media` and density breakpoints.
I haven't personally used any of this before, so I could be misunderstanding, but from what I see online this is supposed to allow the browser to make informed decisions, and also saves you the complexity of dealing with multiple sources + srcset options (of course, multiple sources will still be needed for webp vs. jpg).

It isn't a huge deal one way or the other, so it's also fine to leave things as-is if you prefer, but if it could lead to simplifying the code, providing crisper images on many devices, while keeping the same assets with the same sizes, I figured it's at least worth considering.

src/app/landing/LandingPage.jsx
```diff
@@ -57,34 +64,90 @@ class LandingPage extends Component {
...
               <h2>Search for what you need</h2>
               <p className="text-lighter">
-                GoGetta has New York’s most up-to-date service information on food, personal care, shelter, and clothing.
+                GoGetta has New York’s most up-to-date service information on
+                food, personal care, shelter, and clothing.
               </p>
             </div>
 
             <picture>
+              {/* desktop jpg */}
```
Rovack, line 73: As I'm sure I've noted in the past, I always consider comments a red flag that something in the code isn't as readable as it could be.

In this case, one easy improvement, that would not only make this clear but also save you duplication and thus make this less error-prone, would be to put `1224` into a variable with an indicative name (e.g. desktopPictureMinWidth).

You could make it even clearer, and save even more duplication, by turning this into a reusable component, that would encapsulate all the specific values etc, and then the prop names would take the part of the comment, e.g.:
```
<ResponsivePicture
  desktopJpgSrcset="/img/landing_page/step1_desktop1x.jpg 1x, /img/landing_page/step1_desktop2x.jpg 2x",
  desktopWebpSrcset="/img/landing_page/step1_desktop1x.webp 1x, /img/landing_page/step1_desktop2x.webp 2x",
  mobileJpgSrcset="..."
  mobileWebpSrcset="..."
  defaultImgSrc="..."
  className="..."
  alt="..."
/>
```

And if you wanna take it really far and eliminate all duplication, you could even do what you did with the other components - create a convention for the file names - and make this as simple as:
```
<ResponsivePicture
  imageBasePath="/img/landing_page/step1"
  className="..."
  alt="..."
/>
```
(This would also guarantee that any changes you want to make in the future, like change the logic of how image versions are selected as discussed in the comment above, would require changing only a single file. You could even use it in the testimonial and partner pictures, for maximum effect.)

None of this is necessary, of course, but even the minimal change of using a variable would've achieved the same level of readability, without the risk of getting out of sync with the code as comments often do, would've eliminated some duplication, and likely wouldn't have taken any longer to add than the comment did.

src/app/landing/LandingPage.jsx
```diff
@@ -57,34 +64,90 @@ class LandingPage extends Component {
...
+                Filter or answer a few questions and we will show you locations
+                that meet your specific needs, from age to gender to family
+                situations.
               </p>
             </div>
-            <img className="OnboardingImage" src="img/landing_page/welcome.jpg" alt="" />
+            <picture>
+              {/* desktop jpg */}
+              <source
+                media="(min-aspect-ratio: 2/3)"
```
Rovack, line 123: How come step 2 still uses ratio as the breakpoint, whereas 1 and 3 use width?

src/app/landing/PartnerPicture.jsx
```diff
@@ -1,12 +1,38 @@
...
 import './PartnerPicture.css';
 
-const PartnerPicture = ({ src, alt, link }) => (
+const PartnerPicture = ({ imagePartialPath, alt = '', link }) => (
   <div className="PartnerPicture">
     <a href={link}>
-      <img src={src} alt={alt} />
+      <picture>
+        <source
+          className="OnboardingImage"
```
Rovack, line 10: I thought className only needed to be set on the `img` element, no?
And if not, don't Testimonial.jsx and all the pictures in LandingPage.jsx also need the className on the `source`s as well?

src/app/landing/PartnerPicture.jsx
```diff
@@ -1,12 +1,38 @@
...
+          media="(min-width: 1224px)"
+          srcSet={`/img/partners/${imagePartialPath}_desktop.webp`}
+        />
+        <source
+          className="OnboardingImage"
+          media="(min-width: 1224px)"
+          srcSet={`/img/partners/${imagePartialPath}_desktop.jpg`}
+        />
+        <img
```
Rovack, line 19: No webp for mobile?

src/app/landing/Testimonial.css
```diff
@@ -33,3 +27,12 @@
 .Testimonial-name {
   font-style: italic;
 }
+
+.Testimonial img {
+  width: 100px;
+  border-radius: 50%;
+  border: 3px solid white;
+  -webkit-box-shadow: 0px 3px 8px 2px rgba(184,184,184,1);
```
Rovack, line 35: Doesn't our webpack config add these anyway in the build once you specify box-shadow?

src/app/landing/Testimonial.jsx
```diff
@@ -1,14 +1,51 @@
...
     <div className="Testimonial-separator" />
-
     <div className="Testimonial-text">“{text}”</div>
     <div className="Testimonial-name">– {name}</div>
   </div>
 );
 
+Testimonial.propTypes = {
+  name: PropTypes.string.isRequired,
+  img: PropTypes.shape({
```
Rovack, line 44: This propType doesn't match the props actually being used.

src/app/landing/LandingPage.css
Vikariusu: Background image overflows on very small devices (like iPhone5/SE) so this line resolves that issue, open to suggestions for improvement!

src/app/landing/LandingPage.css
Vikariusu: Oops, fixed!

src/app/landing/LandingPage.css
Vikariusu: There was a weird issue so I left it in but it's fixed now so I'm removing the redundant code

src/app/landing/LandingPage.css
Vikariusu: Fixed!

src/app/landing/LandingPage.jsx
```diff
@@ -57,34 +64,90 @@ class LandingPage extends Component {
...
-                srcSet="img/landing_page/map_tablet.png"
+                media="(min-width: 1224px)"
+                srcSet="/img/landing_page/step1_desktop1x.jpg 1x,
+                /img/landing_page/step1_desktop2x.jpg 2x"
+                type="image/jpeg"
+              />
+              {/* desktop webp */}
+              <source
+                media="(min-width: 1224px)"
+                srcSet="/img/landing_page/step1_desktop1x.webp 1x,
```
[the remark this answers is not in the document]
Vikariusu: Fixed!

src/app/landing/LandingPage.jsx
```diff
@@ -57,34 +64,90 @@ class LandingPage extends Component {
...
+              /img/landing_page/step1_desktop2x.webp 2x"
+                type="image/webp"
+              />
+              {/* mobile jpg */}
+              <source
+                media="(max-width: 1223px)"
```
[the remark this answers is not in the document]
Vikariusu: Fixed

src/app/landing/LandingPage.css
```diff
@@ -249,3 +254,26 @@
     font-size: 11px;
   }
 }
+
+@media screen and (max-width: 1224px) {
+  .LandingPage .IntroBody {
+    background: linear-gradient(to bottom, rgba(0, 0, 0, 0), rgba(0, 0, 0, 0.6)),url('/img/landing_page/top1_mobile.jpg') no-repeat center center fixed;
```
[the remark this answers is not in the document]
Vikariusu: There were 3 JPGs but the highest resolution is almost 8MB in size. I can add it back if we want to use it. I think these are sufficiently small and high quality.

src/app/landing/LandingPage.jsx
Vikariusu: I used Di’s retina mobile images and further compressed the jpgs/ webps so that they would be small enough for use on both kinds of devices and high quality enough that users (hopefully) shouldn’t notice unless they’re looking very closely.

src/app/landing/LandingPage.jsx
Vikariusu: Took out comments and did a little cleanup by setting up const for desktop pixel size

src/app/landing/LandingPage.jsx
Vikariusu: Fixed

src/app/landing/PartnerPicture.jsx
Vikariusu: Yeah, they’re only needed for img elements, removed

src/app/landing/PartnerPicture.jsx
Vikariusu: Added!

src/app/landing/Testimonial.css
Vikariusu: I think it should, there is autoprefixer package in the nodule_modules so I’m removing this line

src/app/landing/Testimonial.jsx
Vikariusu: Good catch, fixed both in Testimonial and PartnerPicture component

src/app/landing/LandingPage.jsx
Rovack: Alright. I suspect for people on a high-end phone it might not feel as sharp as it could, but that's not really the demographic we care about, so that's fine.

src/app/landing/PartnerPicture.jsx
Rovack: Wait, but also, don't you have to specify `type` for webp images? Otherwise I'm not sure this will work on Safari.
Also, in other places I see you had the mobile jpg both as a `source` and `img`. Here you don't need it?

Rovack: (Incidentally, this is why it generally helps to have a single component take care of all instances of logic like this. It avoids the risk of forgetting a version in 1 place, or having slightly different logic.)

PR #168, opened 2020-05-25
Add new fields to Service

abimaelmartell:
Added new fields to service form.

Membership criteria
Service area

Food
- PLHIV Nutrition Program

Clothing
- Gender and Age
- Occasions
- TGNC Affirmative

### DIFF
--- a/src/app/team/service/details/ServiceDetails.jsx
+++ b/src/app/team/service/details/ServiceDetails.jsx
@@ -24,15 +24,28 @@ const LoadingView = () => (
   </div>
 );
 
-function getUpdatedAt(service, metaDataSection, fieldName) {
+function getUpdatedAt(service, route) {
+  const { metaDataSection, fieldName, taxonomySpecificFieldName, selector } = route;
+
+  if (selector) {
+    const value = selector(service);
+    return value ? value.updated_at : null;
+  }
+
+  if (taxonomySpecificFieldName) {
+    const attributes = service.ServiceTaxonomySpecificAttributes;
+    const field = attributes.find(el => el.attribute.name === taxonomySpecificFieldName);
+    return field ? field.updated_at : null;
+  }
+
   const subFields = service.metadata[metaDataSection];
   const field = subFields.find(el => el.field_name === fieldName);
   return field ? field.last_action_date : null;
 }
 
 function ListItem({ route, linkTo, service }) {
-  const { label, metaDataSection, fieldName } = route;
-  const updatedAt = getUpdatedAt(service, metaDataSection, fieldName);
+  const { label } = route;
+  const updatedAt = getUpdatedAt(service, route);
   return <FieldItem title={label} linkTo={linkTo} updatedAt={updatedAt} />;
 }
 
--- a/src/app/team/service/details/ServiceDetailsContainer.jsx
+++ b/src/app/team/service/details/ServiceDetailsContainer.jsx
@@ -6,7 +6,7 @@ import {
   selectLocationError,
   selectLocationData,
 } from '../../../../selectors/location';
-import { getService } from '../../../../selectors/service';
+import { getService, getServiceTaxonomy } from '../../../../selectors/service';
 import * as actions from '../../../../actions';
 
 import ServiceDetails from './ServiceDetails';
@@ -31,6 +31,16 @@ class ServiceDetailsContainer extends Component {
     return `/team/location/${locationId}/services/${serviceId}`;
   };
 
+  filteredServiceFields = () => (
+    SERVICE_FIELDS.filter((el) => {
+      if (el.serviceTaxonomy) {
+        return el.serviceTaxonomy === this.props.serviceTaxonomy;
+      }
+
+      return true;
+    })
+  );
+
   render() {
     return (
       <ServiceDetails
@@ -39,7 +49,7 @@ class ServiceDetailsContainer extends Component {
         service={this.props.service}
         locationData={this.props.locationData}
         locationError={this.props.locationError}
-        serviceFields={SERVICE_FIELDS}
+        serviceFields={this.filteredServiceFields()}
       >
         <Button fluid primary onClick={this.onGoToDocs}>
           Go to docs required
@@ -53,6 +63,7 @@ const mapStateToProps = (state, ownProps) => ({
   service: getService(state, ownProps),
   locationData: selectLocationData(state, ownProps),
   locationError: selectLocationError(state, ownProps),
+  serviceTaxonomy: getServiceTaxonomy(state, ownProps),
 });
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
--- a/src/app/team/serviceForm/area/Edit.jsx
+++ b/src/app/team/serviceForm/area/Edit.jsx
@@ -1,0 +1,166 @@
+import React, { Component, createRef } from 'react';
+
+import Header from '../../../../components/header';
+import Button from '../../../../components/button';
+import Selector from '../../../../components/selector';
+
+import ZipCodeForm from './ZipCodeForm';
+
+import './AreaEdit.css';
+import { AREA_TYPE_LABELS, SERVICE_AREA_TYPES } from './utils';
+
+class WhoDoesItServe extends Component {
+  constructor(props) {
+    super(props);
+
+    let areaType;
+    let customZipCodes;
+
+    if (props.value) {
+      customZipCodes = props.value.postal_codes;
+      areaType = customZipCodes.length === 0 ? 'ALL' : 'CUSTOM';
+    } else {
+      customZipCodes = [];
+      areaType = '';
+    }
+
+    this.state = {
+      areaType,
+      customZipCodes,
+    };
+
+    this.customOcassionRef = createRef();
+    this.updateValue = this.updateValue.bind(this);
+    this.addNewZipCode = this.addNewZipCode.bind(this);
+    this.selectAreaType = this.selectAreaType.bind(this);
+  }
+
+  onDeleteZipCode(index) {
+    const { customZipCodes } = this.state;
+
+    customZipCodes.splice(index, 1);
+
+    this.setState({ customZipCodes });
+  }
+
+  onUpdateZipCode(index, value) {
+    const { customZipCodes } = this.state;
+
+    customZipCodes[index] = value;
+
+    this.setState({
+      customZipCodes,
+    });
+  }
+
+  addNewZipCode() {
+    const customZipCodes = [
+      ...this.state.customZipCodes,
+      '',
+    ];
+
+    this.setState({
+      customZipCodes,
+    });
+  }
+
+  selectAreaType(areaType) {
+    const state = { areaType };
+
+    if (
+      areaType === SERVICE_AREA_TYPES.CUSTOM &&
+      this.state.customZipCodes.length === 0
+    ) {
+      state.customZipCodes = [''];
+    }
+
+    this.setState(state);
+  }
+
+  updateValue() {
+    let { customZipCodes } = this.state;
+
+    if (this.state.areaType === SERVICE_AREA_TYPES.ALL) {
+      customZipCodes = [];
+    }
+
+    this.props.updateValue(
+      {
+        postal_codes: customZipCodes,
+      },
+      this.props.id,
+      this.props.metaDataSection,
+      this.props.fieldName,
+    );
+  }
+
+  render() {
+    const { state: { customZipCodes, areaType } } = this;
+
+    const lastZipCode = customZipCodes[customZipCodes.length - 1];
+
+    return (
+      <div className="w-100 WhoDoesItServe">
+        <Header className="mb-3">Which area does it serve?</Header>
+        <p>(select all that applies)</p>
+        <Selector fluid>
+          <Selector.Option
+            disablePadding={false}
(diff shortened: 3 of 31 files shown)


src/app/team/serviceForm/whoIsThisClothingFor/ServiceWhoIsThisClothingFor.jsx
```diff
@@ -0,0 +1,36 @@
...
+  resourceLoadError: selectLocationError(state, ownProps),
+});
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  fetchResourceData: bindActionCreators(actions.getLocation, dispatch),
+  updateValue: (wearerAge, serviceId, metaDataSection, fieldName) =>
+    dispatch(actions.updateService({
+      locationId: ownProps.match.params.locationId,
+      serviceId,
+      params: { wearerAge },
```
Rovack, line 30: Hmm... this may not be clear from the designs, but I think wearerAge isn't necessarily the only parameter being set here. Looking at the existing data in our DB, at the Airtable I shared a while back, or at the way our filters work right now, you can see that wearerAge actually only supports the values "adults" and "children", whereas the difference between men vs. women or boys vs. girls is currently captured by the "gender" eligibility parameter.
Of course, this is problematic, because it doesn't actually make it possible to capture the _connection_ between age and gender (e.g. there's no way to capture that a place has clothes only for women and boys).

What I'd suggest for now, so that you don't have to change the filters, but also don't have to make any schema changes, is this:
* Keep passing the values as you are right now, for "men", "women", "girls", and "boys".
* When you pass "men" and/or "women", also _add_ the value "adults". And when you pass "boys" and/or "girls", also add the value "children".
* Finally, when "men" is one of the values, also pass the "gender" eligibility parameter with "male", and when "women" is one of the values, pass "female".

That way, we're capturing all the granular data, but we're also backwards-compatible for the values expected by the filters.

src/app/team/serviceForm/whoIsThisClothingFor/WhatKindOfClothingView.jsx
```diff
@@ -0,0 +1,38 @@
...
+import PropTypes from 'prop-types';
+
+import Selector from '../../../../components/selector';
+import ConfirmationOptions from '../../../../components/form/ConfirmationOptions';
+
+function FormView({ value, onConfirm, onEdit }) {
+  return (
+    <div className="w-100">
+      <div style={{ fontSize: '13px', marginBottom: '1em' }} className="font-weight-bold mt-2">
+        What kind of clothing is it?
```
Rovack, line 11: Shouldn't this be `Who is the clothing for?`?
(The filenames - WhatKindOfClothingView, WhatKindOfClothingEdit - also make this confusing.)

src/reducers/locations.js
```diff
@@ -83,6 +91,60 @@ const locationsReducer = (state = {}, action) => {
...
 
+        if (area != null) {
+          ServiceAreas[0] = {
+            ...area,
+            updated_at: dateString,
+          };
+        }
+
+        if (membership) {
+          const membershipIdx = Eligibilities
```
Rovack, line 102: Nitpick, but any reason this isn't just a `find`?

src/reducers/locations.js
```diff
@@ -83,6 +91,60 @@ const locationsReducer = (state = {}, action) => {
...
+        }
+
+        const taxonomySpecificAttributes = [
+          'hasHivNutrition',
+          'tgncClothing',
+          'clothingOccasion',
+          'wearerAge',
+        ];
+
+        taxonomySpecificAttributes.forEach((attr) => {
```
Rovack, line 128: Nice. Cool how you avoided duplication here.

src/reducers/locations.js
abimaelmartell: Because we pass the `Eligibilities` object at the end, so  if we use find, we will have to construct a new array

src/components/input/Input.jsx
```diff
@@ -41,6 +41,7 @@ class Input extends Component {
       onBlur,
       onKeyPress,
       innerRef,
+      id
```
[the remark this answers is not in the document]
abimaelmartell: I'm just picky sometimes to sync label with input, and this can only be done using `id` on the Input, and `htmlFor` on the label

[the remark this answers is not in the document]
Rovack: Makes sense.

src/reducers/locations.js
Rovack: Not sure why that would be needed. I mean, since objects are always just referenced in JS (including when they're in an array) and `find` doesn't create a copy of the object, it should work either way.

I mean:
```
> const a = [{ a: 1, b: 2 }, { a: 2 }]
> const foo = a.find(x => x.a === 2)
> foo.b = 3;
> a
[ { a: 1, b: 2 }, { a: 2, b: 3 } ]
```

But maybe I misunderstood your comment.

src/app/team/serviceForm/area/Edit.jsx
```diff
@@ -0,0 +1,166 @@
...
+    this.customOcassionRef = createRef();
+    this.updateValue = this.updateValue.bind(this);
+    this.addNewZipCode = this.addNewZipCode.bind(this);
+    this.selectAreaType = this.selectAreaType.bind(this);
+  }
+
+  onDeleteZipCode(index) {
+    const { customZipCodes } = this.state;
+
+    customZipCodes.splice(index, 1);
```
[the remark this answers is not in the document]
Rovack: I see onUpdateZipCode still mutates component state, but oh well. Hopefully it won't cause any major issues.

src/app/team/serviceForm/area/Edit.jsx
```diff
@@ -0,0 +1,166 @@
...
+  updateValue() {
+    let { customZipCodes } = this.state;
+
+    if (this.state.areaType === SERVICE_AREA_TYPES.ALL) {
+      customZipCodes = [];
+    }
+
+    this.props.updateValue(
+      {
+        postal_codes: customZipCodes,
```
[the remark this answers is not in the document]
Rovack: Looks like updateValue was indeed updated, but `value` is still a wrapping object. But this was a nitpick anyway, so alright.

src/app/team/serviceForm/whoIsThisClothingFor/options.js
```diff
@@ -0,0 +1,34 @@
...
+  },
+  {
+    label: 'Women',
+    wearerAge: ['women', 'adults'],
+    gender: 'female',
+  },
+  {
+    label: 'Girls',
+    wearerAge: ['girls', 'children'],
+    gender: 'female',
```
Rovack, line 15: Gender should only be added for Men and Women but not Boys and Girls, as otherwise with the current filters, a user that's an adult male could be referred to a location that has clothes only for Women and Boys.

src/app/team/serviceForm/whoIsThisClothingFor/ServiceWhoIsThisClothingFor.jsx
Rovack: Great implementation.

src/app/team/service/details/ServiceDetails.jsx
```diff
@@ -24,15 +24,28 @@ const LoadingView = () => (
   </div>
 );
 
-function getUpdatedAt(service, metaDataSection, fieldName) {
+function getUpdatedAt(service, route) {
+  const { metaDataSection, fieldName, taxonomySpecificFieldName, selector } = route;
```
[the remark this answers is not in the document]
Rovack: Oh, so this really did just end up working out of the box without needing any changes? Cool.

src/app/team/serviceForm/whoIsThisClothingFor/ServiceWhoIsThisClothingFor.jsx
```diff
@@ -0,0 +1,42 @@
...
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  fetchResourceData: bindActionCreators(actions.getLocation, dispatch),
+  updateValue: (value, serviceId, metaDataSection, fieldName) =>
+    dispatch(actions.updateService({
+      locationId: ownProps.match.params.locationId,
+      serviceId,
+      params: {
+        wearerAge: getWearerAgeFromOptions(value),
+        gender: {
+          eligible_values: getGenderFromOptions(value),
```
Rovack, line 34: How come gender needs eligible_values in an object whereas wearerAge directly takes the array? Is it just some difference in the API between eligibility and taxonomy-specific attributes or something?

Rovack:
Awesome, looks like the majority of comments have been addressed, and so I've resolved their conversations.
And nearly all of the unresolved comments are questions, not necessarily actual changes needed.

I think the only one that may be a non-trivial change is how membership criteria still assumes values are unique whereas in reality 2 of the options have the same values, which means there might be a bug there.

Finally, re-iterating my comment about the PR as a whole, it would be great to have some screenshots or clip or something demonstrating how this works.
Thanks!

src/app/team/serviceForm/whoIsThisClothingFor/ServiceWhoIsThisClothingFor.jsx
abimaelmartell: Yes, thats how the api works.

src/app/team/serviceForm/membershipCriteria/options.js
```diff
@@ -4,12 +4,7 @@ const options = [
     values: [true, false],
   },
   {
-    label: 'Referal Letter',
```
Rovack: The commit that adds this to the Proofs page will be in the other PR, I assume?

src/app/team/service/details/ServiceDetailsContainer.jsx
```diff
@@ -31,6 +31,16 @@ class ServiceDetailsContainer extends Component {
     return `/team/location/${locationId}/services/${serviceId}`;
   };
 
+  filteredServiceFields = () => (
+    SERVICE_FIELDS.filter((el) => {
```
Rovack, line 35: I just noticed 1 other bug: This does filter out the irrelenvat fields from the _list_, but when a field is done and the next one needs to be chosen, there's no filtering.
So, for instance, I can do the following:
1. Go into a Food service
2. Click into the "PLHIV" field
3. Enter a value, click "OK"
4. Click "Yes" to confirm

What I expect is to go to "Membership", which is the next field on the list. But what actually happens is that I get to "Occasions", which doesn't even exist for Food services. (Same thing happens if I just use the down/up arrows to navigate between the fields.)

src/app/team/serviceForm/tgncClothing/ServiceTgncClothing.jsx
```diff
@@ -0,0 +1,38 @@
...
+const mapStateToProps = (state, ownProps) => ({
+  resourceData: getService(state, ownProps),
+  value: getServiceTgncClothing(state, ownProps),
+  id: getServiceId(ownProps),
+  resourceLoadError: selectLocationError(state, ownProps),
+});
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  fetchResourceData: bindActionCreators(actions.getLocation, dispatch),
+  updateValue: (tgncClothing, serviceId, metaDataSection, fieldName) =>
```
Rovack, line 26: Is it possible there's some problem with this update? I was trying it locally, and got an error whenever I tried to save any value here (something about "id of undefined" in services/service.js, on the api side).
But maybe that's just a local issue, or maybe it's one of those you've already fixed.

abimaelmartell: I remember on local i was missing a field, this must be it.

Is it possible that the migration didn't add the `tgncClothing` field on staging?

src/app/team/service/details/ServiceDetailsContainer.jsx
abimaelmartell: Yes, you are right, i just fixed this

src/app/team/serviceForm/tgncClothing/ServiceTgncClothing.jsx
Rovack: Just checked, and Staging does have `tgncClothing` in taxonomy_specific_attributes.

abimaelmartell: I have two DB dumps on my local files, i tried both and there is one where the `tgncClothing` field does not exist, can you confirm that on your local setup that field exists?

src/app/team/serviceForm/membershipCriteria/options.js
Rovack: @abimaelmartell I didn't see the "Referral Letter" option in the proofs PR.
Shouldn't that be added to the proofs page, since it was removed from here?

src/app/team/serviceForm/tgncClothing/ServiceTgncClothing.jsx
Rovack: My local DB doesn't have `tgncClothing`, though I'm pretty sure I saw this problem when working with the Staging DB, not my local one.

But if you're confident that's the only cause for this error, then that's OK.
We can merge this, deploy to Staging, and hope it works.

PR #171, opened 2020-05-29
fix: remove no results message on initial load

Vikariusu:


Rovack:
So, this change does make the "No results found" flash basically imperceptible, but there are a couple of issues with it:
1. As you can see by adding a few console logs, the "no results" does still get rendered and may potentially flash, it's just extremely quick now
2. The reason it isn't as noticeable now is actually the introduction of a "bug": When searchLocations gets called and sets isSearchingLocations to true, but then fetchLocations resolves immediately because of the `if (!center || !categories)` check, isSearchingLocations just stays true even though no search is actually happening. The only reason it doesn't keep saying "Loading results..." forever is that another search starts very quickly after that, but there could be cases where no additional search gets triggered, and it actually does get stuck that way. If that does happen, it's actually worse than the bug we're trying to fix

Given this, I think a fix that would completely eliminate the flash, and wouldn't risk any new bugs, would be moving/copying the `if (!center || !categories)` check to searchLocations.

(Honestly, the need for that check in itself is kind of confusing, and ideally we'd figure out a way that it wouldn't be needed at all, but that might be a more complicated change.)

Rovack:
Looks good. Just, as mentioned, we should move setting isSearchingLocations to false back to searchLocations, to avoid a potential infinite-loading bug.

Rovack:
> Looks good. Just, as mentioned, we should move setting isSearchingLocations to false back to searchLocations, to avoid a potential infinite-loading bug.

Rovack:
Whoops, accidentally clicked the Close button... 😶 

Anyway, looking at the last commit, I think there might have been a misunderstanding: I just meant moving the `isSearchingLocations: false` setState back from fetchLocations into searchLocations, due to the potential edge case I mentioned in my first comment. Basically, that just means reverting the first commit on this PR.
Not sure what purpose this latest commit serves.

Maybe we should just talk for a couple minutes after standup if there's any unclarity about the issue to which I'm referring.

Vikariusu:
Reverts completed, I tested the code and did not encounter any premature no results errors (tested for flashing and by putting in console.logs).

PR #172, opened 2020-06-04
fix/replace-hours-with-midnight

cadecameron:
Small PR to fix [Replace 11:59 and 00:00 with "Midnight"](https://app.asana.com/0/1168502694533225/1175729283540779/f)

Screenshot:
Screen Shot 2020-06-04 at 12 29 40 PM

Time schedule related to screenshot:
Screen Shot 2020-06-04 at 12 30 01 PM

### DIFF
--- a/src/app/find/locationDetails/ServiceSection.jsx
+++ b/src/app/find/locationDetails/ServiceSection.jsx
@@ -15,7 +15,11 @@ const renderSchedule = (schedule) => {
 
   const dayNumberToName = weekday => DAYS[weekday - 1];
 
-  const formatHour = time => moment(time, 'HH:mm:ss').format('LT').replace(':00 ', ' ');
+  const formatHour = time => moment(time, 'HH:mm:ss')
+    .format('LT')
+    .replace(':00 ', ' ')
+    .replace(/(11:59 PM)|(12 AM)/, 'midnight');
+
   const formatHours = (opens, closes) => `${formatHour(opens)} to ${formatHour(closes)}`;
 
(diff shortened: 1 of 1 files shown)


src/app/find/locationDetails/ServiceSection.jsx
```diff
@@ -15,7 +15,11 @@ const renderSchedule = (schedule) => {
 
   const dayNumberToName = weekday => DAYS[weekday - 1];
 
-  const formatHour = time => moment(time, 'HH:mm:ss').format('LT').replace(':00 ', ' ');
+  const formatHour = time => moment(time, 'HH:mm:ss')
+    .format('LT')
+    .replace(':00 ', ' ')
+    .replace(/(11:59 PM)|(12 AM)/, 'midnight');
```
Rovack, line 21: Would the format always be exactly this one? IIRC `LT` it a locale-aware format, meaning it would show different strings on different devices.
If that is in fact the case, I'd suggest doing the conversion to midnight while it's still a value, before the formatting turns it into a string.

cadecameron: Oh, good point. I didn't realize it was locale-aware.

cadecameron:
Let me know if this is more appropriate @Rovack, or if you can think of any other edge cases.

cadecameron:
Actually, when you actually look at the data for this location in the image above, it would seem that it's possibly been input incorrectly? This data comes straight from the test deployment on AWS.

PR #173, opened 2020-06-05
fix/split-phone-numbers

cadecameron:
@Rovack 

I think these were the correct lines to remove, though you might want to double check. If you have a location that would be better suited for testing this functionality, please let me know it's ID/name.

Resulting phone number list with lines removed:
Screen Shot 2020-06-05 at 3 25 14 PM

Phone data related to this screenshot:
Screen Shot 2020-06-05 at 3 25 40 PM

### DIFF
--- a/src/app/find/locationDetails/LocationDetails.jsx
+++ b/src/app/find/locationDetails/LocationDetails.jsx
@@ -123,9 +123,6 @@ const renderLocation = (location, searchCategory) => {
       service.metadata.sources.some(source => source.includes('FPC')));
 
   const phones = [];
-  if (location.Organization.Phones) {
-    phones.push(...location.Organization.Phones);
-  }
   if (location.Phones) {
     location.Phones.forEach((phone) => {
       if (!phones.find(existingPhone => existingPhone.number === phone.number)) {


Rovack:
Huh, so this has only been merging _location_ and _organization_ phone numbers?

When the task was created, it was because Adam was saying the phones for individual _services_ were being displayed along with the location numbers, but looking at this code I see that's not the case.

Well, in that case, I'm not sure there's a problem at all. 😶
Maybe ask Adam if he can find an example of this problem, but if he can't, then I guess this task should just be closed.

cadecameron:
There seems to be multiple phone number sources: `organization`, `location`, and you're suggesting a third: `services`?

I'll take a look at the data on Monday and see what's going on then.

Rovack:
Yup, that's exactly right.

cadecameron:
> When the task was created, it was because Adam was saying the phones for individual _services_ were being displayed along with the location numbers, but looking at this code I see that's not the case.

I have a feeling he was referring to the `location.Organization.Phones` data as 'services'. As far as I can tell, the application currently doesn't render phone numbers found in `location.Services[0].Phones` anywhere (not that there's many in the database).

If that's the case, then this code deletion should do the trick.

PR #175, opened 2020-06-28
Fix safari search

AdamNgu:
Tested and confirmed working on: Safari, Chrome, and XCode iPhone simulator


Rovack:
@AdamNgu One thing I did just notice is that this was merged into feature/coronavirus-sstt, though it will be needed after Covid as well. So please cherry-pick it into Master if you can.

PR #177, opened 2020-07-12
feat: add basic mixpanel support

Vikariusu:
This adds a basic reusable mixpanel function to track user events with all the basic features mixpanel needs.

Note: before merging, we need to set up our .env files in both dev and prod environments with the API key for mixpanel.

### DIFF
--- a/src/app/App.jsx
+++ b/src/app/App.jsx
@@ -18,13 +18,15 @@ import NotFound from './notFound/NotFound';
 import LoadingLabel from '../components/form/LoadingLabel';
 
 import withTracker from '../components/routing/withTracker';
+import analytics from '../services/analytics';
 
 const TeamRouter = React.lazy(() => import('./team/Router'));
 const CommentsRouter = React.lazy(() => import('./comments/Router'));
 const findRouterPromise = import('./find/Router');
 const FindRouter = React.lazy(() => findRouterPromise);
 
-history.listen((location, action) => {
+history.listen(() => {
+  analytics.track('Page View');
   window.scrollTo(0, 0);
 });
 
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 
 # misc
 .DS_Store
+.env
 .env.local
 .env.development.local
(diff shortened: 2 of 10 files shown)


Rovack:
Nice!

Just 2 notes:
* I think it'd be nice to rename the service from Mixpanel to just "analytics" or "events" or something. I mean, the function params are general enough to likely to work for any analyrics product, and if we e.g. switch to Amplitude, that way we'd only need to change 1 file.
* The custom events that can be added wherever they're needed are great, but can we also have something that automatically tracks every navigation, like we do for Google Analytics? That will make it easier for anyone wtih access to Mixpanel to set up tracking of whatever funnels they want, without us having to deploy a version for every change. You could probably do it by just integrating with withTracker.js, or maybe even with react-router if there's some package for connecting mixpanel-browser with it.

src/app/App.jsx
```diff
@@ -18,13 +18,15 @@ import NotFound from './notFound/NotFound';
...
+import analytics from '../services/analytics';
 
 const TeamRouter = React.lazy(() => import('./team/Router'));
 const CommentsRouter = React.lazy(() => import('./comments/Router'));
 const findRouterPromise = import('./find/Router');
 const FindRouter = React.lazy(() => findRouterPromise);
 
-history.listen((location, action) => {
+history.listen(() => {
+  analytics.track('Page View');
```
Rovack, line 29: Ah, awesome!
Can we make the event params include the URL and such, so we can know which page it is?

Vikariusu: Properties called "Current URL" (https://gogetta.nyc/find) and "url" (/find) are automatically added by Mixpanel :)

Rovack: Ah, great. 😄

Rovack:
All comments have been addressed.
Should we merge this now, or wait until the env var is passed somehow, so it can actually send to Mixpanel?

Vikariusu:
I'll do more research on that first + add some condition to make sure that not having the key set won't break the code.

PR #180, opened 2020-07-31
Resolve issue where icon is too far to right on iphone 4

AdamNgu:


src/app/find/locationDetails/LocationDetails.jsx
```diff
@@ -280,6 +280,7 @@ class LocationDetails extends Component {
           className="locationCloseButton"
           name="times"
           onClick={goBack}
+          style={{ marginRight: '10px' }}
```
Rovack, line 283: The styles for `locationCloseButton` are in locationDetails.css.

Rovack: And btw, I figure increasing the `right` value there might have the same effect as adding `margin-right`.

Rovack:
Just to make sure, because we've had issues with this before: Does this work with long titles, or is there a risk of the "x" clashing with the text?
If there is a risk, consider decreasing the width percentage of `.locationTitle`.

PR #181, opened 2020-08-04
feat: add a confirmation button to FiltersModal component

Vikariusu:


src/app/find/mapPage/filters/filters.css
```diff
@@ -29,3 +29,13 @@ input#hivNutrition-checkbox {
 	margin-left: 32px;
 	color: #333333;
 }
+
+.filterSubmitBtn {
+    position: fixed;
```
Rovack, line 34: Indentation is inconsistent with the rest of the file.

Rovack, line 34: One issue I've often seen with fixed-position bottom elements: If the contents of the screen take up 100vh or more, often the bottom-most elements can never be seen properly, since scrolling down all the way will leave the floating element covering them.
One way around that is to add sufficient bottom padding to the whole container.

Is this an issue here, or is there already sufficient empty space at the bottom to accommodate the button?

src/app/find/mapPage/filters/filters.css
```diff
@@ -29,3 +29,13 @@ input#hivNutrition-checkbox {
...
+    padding: 16px;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    z-index: 2;
+    background-color: rgba(255, 255, 255, 0.82);
```
Rovack, line 40: This isn't one of the greys we already have in colors.js (or very similar to one of them)?

src/app/find/mapPage/filters/FiltersModal.jsx
```diff
@@ -51,6 +52,11 @@ class FiltersModal extends Component {
             onChange={this.setFilterValues}
           />
         </div>
+        <div className="d-flex justify-content-around filterSubmitBtn">
```
Rovack, line 55: How come flex is needed for a div with only a single child? If it's meant to vertically center the button in the div, then wouldn't that happen anyway, since the spacing is just coming from padding that's equal on top and bottom?

src/app/find/mapPage/filters/FiltersModal.jsx
```diff
@@ -51,6 +52,11 @@ class FiltersModal extends Component {
...
+          <Button secondary fluid onClick={this.submit} >
+            <span>Apply filters</span>
```
Rovack, line 57: Not that it hurts, but does the `span` do anything? (I mean, as opposed to just putting the text as the direct child of the Button.)

src/app/find/mapPage/filters/FiltersModal.jsx
Vikariusu: Fixed!

src/app/find/mapPage/filters/FiltersModal.jsx
Vikariusu: Fixed!

src/app/find/mapPage/filters/filters.css
Vikariusu: The selector above was using tabs, fixed it to match the others.

Vikariusu: Correct, the pb-5 should solve that. I tested it on multiple different screen sizes (including very old phones like Nexus 4 and iPhone 4) and noticed no problems so it should be sufficient.

PR #182, opened 2020-08-04
feature/social-media-icons

cadecameron:
Hi @Rovack ,

I've added our social media icons + links to the bottom of the landing page.
This relates to the Asana task: [Add social media links to landing page](https://app.asana.com/0/1168502694533225/1185125349291226/f).

For the life of me, I couldn't use the `<Icon>` component to get these icons to render (like we do with all the other icons in the site). I'm guessing it's got something to do with the `name` prop (I tried all different names). I ended up importing the `FontAwesomeIcon` component and using it to render instead (since the documentation refers to an `icon` prop, rather than `name`).

### DIFF
--- a/src/app/landing/LandingPage.css
+++ b/src/app/landing/LandingPage.css
@@ -176,16 +176,14 @@
   justify-content: center;
 }
 
-.LandingPage .Feedback .SocialIconsList {
-  margin-top: -5%;
-  margin-left: -3px;
+.LandingPage .Feedback .SocialIcons {
+  margin: -22px 36px 46px;
   font-size: 1.1rem;
 }
 
-.LandingPage .Feedback .SocialIconsList li {
-  display: inline;
-  list-style-type:none;
-  padding-right: 10px;
+.LandingPage .Feedback .SocialIcons a {
+  color: #FFFFFF;
+  padding-right: 8px;
 }
 
 .LandingPage h1 {
--- a/src/app/landing/LandingPage.css
+++ b/src/app/landing/LandingPage.css
@@ -176,6 +176,18 @@
   justify-content: center;
 }
 
+.LandingPage .Feedback .SocialIconsList {
+  margin-top: -5%;
+  margin-left: -3px;
+  font-size: 1.1rem;
+}
+
+.LandingPage .Feedback .SocialIconsList li {
+  display: inline;
+  list-style-type:none;
+  padding-right: 10px;
+}
+
 .LandingPage h1 {
   font-weight: bold;
   margin-left: 25px;
--- a/src/app/landing/LandingPage.css
+++ b/src/app/landing/LandingPage.css
@@ -166,6 +166,14 @@
   text-align: left;
 }
 
+.LandingPage .Feedback .SectionInner h2 {
+  margin-bottom: 6px;
+}
+
+.LandingPage .Feedback .SectionInner p {
+  margin-bottom: 16px;
+}
+
 .LandingPage .Feedback .FeedbackEmail {
   background-color: #FFFFFF;
   margin-left: 36px;
@@ -177,7 +185,7 @@
 }
 
 .LandingPage .Feedback .SocialIcons {
-  margin: -22px 36px 46px;
+  margin: 0px 36px 46px;
   font-size: 1.1rem;
 }
 
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "@fortawesome/fontawesome-svg-core": "^1.2.28",
+    "@fortawesome/free-brands-svg-icons": "^5.14.0",
     "@fortawesome/free-solid-svg-icons": "^5.13.0",
     "@fortawesome/react-fontawesome": "^0.1.9",
     "aws-amplify": "^1.1.6",
--- a/src/app/IconLibrary.js
+++ b/src/app/IconLibrary.js
@@ -31,6 +31,11 @@ import {
   faSpinner,
   faTrash,
 } from '@fortawesome/free-solid-svg-icons';
+import {
+  faTwitterSquare,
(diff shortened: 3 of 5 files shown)


cadecameron:
image

Rovack:
You can see in line 35 of Icon.jsx that it passes the `name` property it's given as the `icon` property to the `FontAwesomeIcon` component.
However, the propTypes does currently specify it's supposed to be a string. If in this case you need to pass an array of strings, you can just update the propTypes.

(I meant to check out this branch to see if that works on my machine, but I see it's in your forked repo.)

Rovack:
Btw, looking at the screenshot, the part about following us and the part about how we're all about feedback feel like they're supposed to be related somehow. Maybe they were further apart in the designs, or there was some other style difference?

src/app/landing/LandingPage.css
```diff
@@ -176,6 +176,18 @@
   justify-content: center;
 }
 
+.LandingPage .Feedback .SocialIconsList {
+  margin-top: -5%;
+  margin-left: -3px;
```
Rovack, line 181: Ugh, negative margins... always a sign of trouble.

Is it because a `ul` is used?
And if so, does it have to be?

cadecameron: > Ugh, negative margins... always a sign of trouble.
> 
> Is it because a `ul` is used?
> And if so, does it have to be?

I restructured the icons so they're simple `<a>` tags now. This helped with the negative left margin, however the `<h2>` tags on the page all have quite a large bottom margin that I needed to reduce.

cadecameron:
> You can see in line 35 of Icon.jsx that it passes the `name` property it's given as the `icon` property to the `FontAwesomeIcon` component.
> However, the propTypes does currently specify it's supposed to be a string. If in this case you need to pass an array of strings, you can just update the propTypes.
> 
> (I meant to check out this branch to see if that works on my machine, but I see it's in your forked repo.)

I've fixed up the prop type in the `Icon.jsx` file to accept arrays of strings (in addition to regular strings).

> Btw, looking at the screenshot, the part about following us and the part about how we're all about feedback feel like they're supposed to be related somehow. Maybe they were further apart in the designs, or there was some other style difference?

I followed the Figma design pretty closely (link in the Asana ticket). I modified the top and bottom margins of the social media icons slightly in this commit.

cadecameron:
New spacing:

image

Rovack:
The code all looks good, although having looked at the designs on Figma, the spacing still seems off: In Figma, it's 91px between the bottom of the icons to the top of "We're all about feedback", whereas the values I see in CSS seem much smaller.

Also, the distance between "We're all about feedback" and "tell us what you think" is much smaller in the designs, which makes it feel like a cohesive unit: in the designs it's 26% of the distance between the social icons and feedback, whereas in the screenshot you attached it's like 60%.

Wouldn't it be possible to just take the exact values used in Figma for all these spaces (+font sizes)?

cadecameron:
> The code all looks good, although having looked at the designs on Figma, the spacing still seems off: In Figma, it's 91px between the bottom of the icons to the top of "We're all about feedback", whereas the values I see in CSS seem much smaller.
> 
> Also, the distance between "We're all about feedback" and "tell us what you think" is much smaller in the designs, which makes it feel like a cohesive unit: in the designs it's 26% of the distance between the social icons and feedback, whereas in the screenshot you attached it's like 60%.
> 
> Wouldn't it be possible to just take the exact values used in Figma for all these spaces (+font sizes)?

I can't see those details in Figma @Rovack ... 🤔
All I can do is pan around the landing page layer.

I've noticed that there's multiple designs that include the social media icons. The one I was referencing for this ticket is this (it has the 'Social Media Icon at the bottom' title:

image

If this is the incorrect design, let me know and I'll update the code with the correct one.

Rovack:
Ohhh, that sounds like the non-logged-in version of Figma.
Di has now invited to that file in Figma, so as long as you're logged in, you should be able to see the distances, sizes, and all the other values. Without those, all a dev can do is eyeball the designs, so it requires more work and ends up with less accurate results.

But anyway, yeah, I was also looking at the part under "Social Media Icon at the bottom", so once you can see the values, the differences should be easy to resolve.

As for the spacing for existing elements ("We're all about feedback", "tell us what you think"...) it's possible it has changed in the designs due to the addition of the social media links to that same footer, and also possible that we just implemented the spacing incorrectly the first time around.
Either way, I suppose it didn't make much difference before, but now that we have 2 separate and unrelated parts in the black footer, each with multiple lines, getting the spacing right becomes more significant, as otherwise it's not as visually clear what the separation is.

cadecameron:
Hey @Rovack,
After signing up and logging into the Figma board, trying to figure out what spacing values to use still isn't exactly clear. Unless there's a dashboard or view I'm not seeing on the Figma site, there's no easy way to figure out margins/spacing between elements?

> The code all looks good, although having looked at the designs on Figma, the spacing still seems off: In Figma, it's 91px between the bottom of the icons to the top of "We're all about feedback", whereas the values I see in CSS seem much smaller.
> 
> Wouldn't it be possible to just take the exact values used in Figma for all these spaces (+font sizes)?

I can see the 91px you referenced here, however if I put a `margin-bottom` of 91px for the `SocialIcons` class, it's _way_ bigger than what the Figma design shows (see image below). 🤷‍♂️

image

For the meantime, I've eyeballed the spacing again, and this is the current product:

image

If I'm missing something obvious with the Figma site, I'd love to do a quick walk-through with you or Di at some stage to demo how to use it to get the correct spacing/dimensions from it.

Rovack:
When I use Figma, whenever I click an element and then hover over another element, it shows me the spacing between them.

But anyway, the side-by-side comparison looks good, and the "follow us" and "feedback" sections feel distinct enough to me, so LGTM.

PR #271, opened 2024-05-11
Integrate sentry for tracking errors on SSTT updates

jbeard4:
Sentry integration is working well.

Sentry dashboard reports request and response:

image

Publish notification to Slack:

image

### DIFF
--- a/src/reducers/dataEntryErrorsReducer.js
+++ b/src/reducers/dataEntryErrorsReducer.js
@@ -1,35 +1,27 @@
+import * as Sentry from '@sentry/browser';
+
 import {
   UPDATE_LOCATION_ERROR,
   UPDATE_SERVICE_ERROR,
   DISMISS_DATA_ENTRY_ERRORS,
 } from '../actions';
-import * as Sentry from '@sentry/browser';
-
-Sentry.init({
-  dsn: "https://f7a71e0f6282e1a79ba6f604080ec88e@o4507232810631168.ingest.us.sentry.io/4507235352313856",
-  // Set 'tracePropagationTargets' to control for which URLs distributed tracing should be enabled
-  tracePropagationTargets: ["localhost", /^https:\/\/(test.)?gogetta.nyc\/team/],
-  // Session Replay
-  replaysSessionSampleRate: 1, 
-  replaysOnErrorSampleRate: 1,
-});
 
 const locationErrorsReducer = (state = [], action) => {
+  const extra = {
+    response: {
+      payload: action.payload.error.response.data,
+      ...action.payload.error.response,
+    },
+  };
   switch (action.type) {
     case UPDATE_LOCATION_ERROR:
     case UPDATE_SERVICE_ERROR:
       if (!action.payload || !action.payload.error) {
         return state;
       }
-      const extra = {
-        response: {
-          payload: action.payload.error.response.data,
-          ...action.payload.error.response,
-        }
-      }
       Sentry.captureException(action.payload.error, {
         contexts: extra,
-        extra: extra,
+        extra,
       });
       return [action.payload.error, ...state];
 
--- a/src/reducers/dataEntryErrorsReducer.js
+++ b/src/reducers/dataEntryErrorsReducer.js
@@ -3,6 +3,16 @@ import {
   UPDATE_SERVICE_ERROR,
   DISMISS_DATA_ENTRY_ERRORS,
 } from '../actions';
+import * as Sentry from '@sentry/browser';
+
+Sentry.init({
+  dsn: "https://f7a71e0f6282e1a79ba6f604080ec88e@o4507232810631168.ingest.us.sentry.io/4507235352313856",
+  // Set 'tracePropagationTargets' to control for which URLs distributed tracing should be enabled
+  tracePropagationTargets: ["localhost", /^https:\/\/(test.)?gogetta.nyc\/team/],
+  // Session Replay
+  replaysSessionSampleRate: 1, 
+  replaysOnErrorSampleRate: 1,
+});
 
 const locationErrorsReducer = (state = [], action) => {
   switch (action.type) {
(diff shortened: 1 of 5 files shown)


src/reducers/dataEntryErrorsReducer.js
```diff
@@ -3,6 +3,16 @@ import {
   UPDATE_SERVICE_ERROR,
   DISMISS_DATA_ENTRY_ERRORS,
 } from '../actions';
+import * as Sentry from '@sentry/browser';
+
+Sentry.init({
+  dsn: "https://f7a71e0f6282e1a79ba6f604080ec88e@o4507232810631168.ingest.us.sentry.io/4507235352313856",
+  // Set 'tracePropagationTargets' to control for which URLs distributed tracing should be enabled
+  tracePropagationTargets: ["localhost", /^https:\/\/(test.)?gogetta.nyc\/team/],
```
Rovack, line 11: IIRC www.gogetta.nyc is also used.
(More generally, any reason to limit the target URLs in the first place? Can't think of any other URLs where this code runs, so is it just that there's no way to specify like `*`?)

src/reducers/dataEntryErrorsReducer.js
```diff
@@ -3,6 +3,16 @@ import {
   UPDATE_SERVICE_ERROR,
   DISMISS_DATA_ENTRY_ERRORS,
 } from '../actions';
+import * as Sentry from '@sentry/browser';
+
+Sentry.init({
+  dsn: "https://f7a71e0f6282e1a79ba6f604080ec88e@o4507232810631168.ingest.us.sentry.io/4507235352313856",
```
Rovack, line 9: Nitpick: Values like this in this repo usually go in config.js.

src/reducers/dataEntryErrorsReducer.js
```diff
@@ -3,6 +3,16 @@ import {
...
+  // Set 'tracePropagationTargets' to control for which URLs distributed tracing should be enabled
+  tracePropagationTargets: ["localhost", /^https:\/\/(test.)?gogetta.nyc\/team/],
+  // Session Replay
```
Rovack, line 12: Is this like FullStory's session recordings? If so, note that with FS I think we had to disable capturing of free text field values, to avoid the risk of anyone entering PII for some reason. Don't know how likely that is, but worth considering.

src/reducers/dataEntryErrorsReducer.js
```diff
@@ -11,7 +21,16 @@ const locationErrorsReducer = (state = [], action) => {
...
         return state;
       }
-
+      const extra = {
+        response: {
+          payload: action.payload.error.response.data,
+          ...action.payload.error.response,
+        }
+      }
+      Sentry.captureException(action.payload.error, {
```
Rovack, line 30: Would it make sense to use Sentry's SDK for [React](https://docs.sentry.io/platforms/javascript/guides/react/), which includes out-of-the-box [Redux support](https://docs.sentry.io/platforms/javascript/guides/react/features/redux/)?

Haven't used it before, so don't know if it would fully avoid the need to explicitly set it up for individual error reducers like this, but if so, it might give more comprehensive coverage with less code.

src/reducers/dataEntryErrorsReducer.js
```diff
@@ -3,6 +3,16 @@ import {
   UPDATE_SERVICE_ERROR,
   DISMISS_DATA_ENTRY_ERRORS,
 } from '../actions';
+import * as Sentry from '@sentry/browser';
+
+Sentry.init({
```
Rovack, line 8: Could be nice to initialize this in its own file under `services`, like we do for e.g. Mixpanel (in `services/analytics.js`), so a specific reducer doesn't have to own the concern of configuring Sentry.

src/reducers/dataEntryErrorsReducer.js
jbeard4: This is a good idea, but let's punt it. For now, we just want to track these HTTP errors.

jbeard4:
Hi @Rovack thanks for your feedback. This is ready for re-review.

src/reducers/dataEntryErrorsReducer.js
jbeard4: Fixed in 5559ebbb6adaaaa09ee40e7a70e02f1cc66b06f6

src/reducers/dataEntryErrorsReducer.js
jbeard4: Fixed in 5559ebbb6adaaaa09ee40e7a70e02f1cc66b06f6

src/reducers/dataEntryErrorsReducer.js
jbeard4: Fixed in 5559ebbb6adaaaa09ee40e7a70e02f1cc66b06f6

src/reducers/dataEntryErrorsReducer.js
jbeard4: Fixed in 5559ebbb6adaaaa09ee40e7a70e02f1cc66b06f6

I'm not sure if there's a reason to limit it. This just seems more correct.

src/reducers/dataEntryErrorsReducer.js
Rovack: Sounds good.

src/services/sentry.js
```diff
@@ -0,0 +1,12 @@
+import * as Sentry from '@sentry/browser';
+
+import config from '../config';
+
+Sentry.init({
+  // Set 'tracePropagationTargets' to control for which URLs distributed tracing should be enabled
+  dsn: config.sentryDsn,
+  tracePropagationTargets: [
+    'localhost',
+    /^https:\/\/(((test)|(www)).)?gogetta.nyc\/team/,
```
Rovack, line 10: One thing I missed is that technically these periods should probably be `\.`, but let's not be _that_ nitpicky. :)

Rovack (approved):

PR #288, opened 2024-11-18
age group

shakilhossain1:


jbeard4:
@shakilhossain1 I deployed this to test environment, but found that it's not working yet:

Screenshot from 2024-11-24 20-08-59

You can see that the types are still incorrect. Sometimes they are still set as strings in the database, so they do not render correctly in YourPeer UI:

```json
[{"name": "Children", "maxAge": 11, "minAge": 0, "allAges": true}, {"name": "my custom group", "maxAge": "22", "minAge": "2", "allAges": false}]
```

shakilhossain1:
This is working now. Can we deploy this?
we need to deploy this [pr](https://github.com/streetlives/streetlives-api/pull/115) also

jbeard4:
I QA's this and found two issues:
* If age_min or age_max is not set, then they should be set to null in the payload. Currently, their values are unset.
* Currently, if "All ages in this group" is selected, then property "all_ages" is set to true in the payload. "all_ages" should be set to true only if "Veterans (all ages)", "Organization clients only (all ages)", or a custom group is selected. Otherwise it should be set to false. And, in the user interface, the way that we know if "All ages in this group" is selected, is if the age_min and age_max properties are equal to the default min and max values for each of the pre-defined groups (e.g. "Children (0-11)", etc.).
image

adambard1 (approved):
Approved

PR #332, opened 2026-06-03
fix: update marker color logic for closed locations

shakilhossain1:


adambard1:
I did some QA on this, and all seems good. Closed locations create grey pins. The closure info screen works as does the hyperlink HTML if we want to redirect users.

